PR #276, opened 2022-03-03
Test that all primitive functions typecheck

georgefst:
We attempt to check that, for every member of `globalPrims`, its declared type matches its behaviour.

This test is unfortunately extremely discard-happy, hence the `withTests 1` currently. Generating complex expressions, then discarding those which are not normal form, is pretty inefficient. _Ideally_ we'd have a generator which produces normal forms. It also always seems to stack overflow on `Int.toNat` for some reason - I haven't looked in to that.

If this is infeasible, it may be better to take a single arbitrary concrete value of that type. But this is far from ideal, since there there can be branches in the implementation of `primFunDef` e.g. for `Int.rem`.

Putting this here as a draft in case anyone has other ideas.

### DIFF
--- a/primer/test/Tests/Typecheck.hs
+++ b/primer/test/Tests/Typecheck.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE TupleSections #-}
 
 -- | Tests for the typechecker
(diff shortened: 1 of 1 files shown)


brprice:
This would (ideally) have caught the bug in #1159 . It does need to be a bit more nuanced than the above "generate a normal form", as `const` is now a primitive that does not require its arguments to be normal.

Having said that, the test is slightly redundant with the `type preservation` test, which did indeed catch the bug. However, a focused test would have caught it sooner, and could have given a much nicer failure report

dhess:
Should we commit this in its current form anyway, then, despite the inefficiencies?

PR #559, opened 2022-07-06
Ensure API calls do not introduce shadowing

brprice:
Both "construction" api calls and evaluation will not introduce shadowing into a program which currently contains none, except in a few specific cases:
- evaluation under a binder in one of two situations (both "inline a global which has a binder of the same name as a local one")
  - inline a global variable, e.g. `λx. foo` reduces to `λx. λx.t` where `foo = λx.t`
    (note that we don't have the same problem for local `let f = λx.t in λx.s` as we will notice the shadowing issue when considering pushing the `let` under the lambda and do an initial renaming step `let f = λx.t in λy.let x=y in t`)
  - case-of-known-constructor, which "inlines" parts of the type definition: `Λα. case C t : T of C x -> s` reduces to `Λα. let x = t : ∀α.A in s` when we have `data T = C (∀α.A)`
- actions
  - anything with a student-specified name, since they can manually request a clashing name
  - converting let to letrec: `let x = (let x = t in ?) in ?` will be converted to `letrec x = (let x = t in ?) in ?` and now the `x` is shadowed in `t`.

To do this, we had to make some minor changes
- TC: forbid `case e of C x x -> ...` where we have duplicate binders. I think that this was never intended to be accepted
- tweak the evaluation of a  `letrec x = t : T in x` from `letrec x = t : T in (t : T)` to `(letrec x = t : T in t) : T`, else `x` can be shadowed in `T`.
- fix a bug in `bindersBelow`
- add an option to eval to avoid shadowing: whenever we may push some lets under a binder we check for shadowing as well as capture; i.e. for `let x = t in λy.s` we always ensured `y` was not free in `t` and `x` is different to `y` (capture),  we now also check that there are no `y` binders inside `t` (shadowing).

### DIFF
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -5,6 +5,7 @@ module Primer.Eval (
   -- The public API of this module
   step,
   redexes,
+  AvoidShadowing (..),
   RunRedexOptions (..),
   ViewRedexOptions (..),
   NormalOrderOptions (..),
@@ -68,7 +69,7 @@ import Primer.Eval.Redex (
   EvalLog (..),
   MonadEval,
   RunRedexOptions (RunRedexOptions, pushAndElide),
-  ViewRedexOptions (ViewRedexOptions, aggressiveElision, groupedLets),
+  ViewRedexOptions (ViewRedexOptions, aggressiveElision, avoidShadowing, groupedLets),
   lookupEnclosingLet,
   runRedex,
   runRedexTy,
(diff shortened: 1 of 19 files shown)


brprice:
I have revisited this work, and we now have test that both eval and offered actions do not introduce shadowing (except in a very few known cases)

primer/src/Primer/Eval.hs
```diff
@@ -130,6 +130,7 @@ evalOpts =
     { groupedLets = True
     , aggressiveElision = True
     , avoidShadowing = True
+    -- TODO: what option do we want here? NB neither is "more permissive", as it swaps "rename" and "push". If change this, will need to think about our test!
     }
```
brprice, line 134: Bikeshed: are people happy with enforcing single-step eval to always avoid shadowing? I could factor this out into a non-hardcoded option if we want.

dhess: Let's leave it as an option, if it's not *too* much extra work.

brprice: It is not much work, it is already an option in the core of eval, just needs piping around through all the relevant functions rather than being picked up from this hardcoded set. (The hardcoded was initially "maximally permissive" for the interactive eval: the student should be able to choose any redex, but since has diverged from that: `aggressiveElision` is somewhat sensible to cut down on "bookkeeping"; `avoidShadowing` should arguably not even be an option, but allow two redexes on the same node "reduce and create shadow" and "rename to avoid later shadow", but we don't have that ability currently).

dhess: The only other objection I'd have is if it slowed down eval, but I think that in single-step mode, it won't matter. This is assuming, of course, that we end up writing a separate fast evaluator for full/run-to-breakpoint evaluation.

georgefst (approved):

primer/test/Tests/Shadowing.hs
```diff
@@ -0,0 +1,327 @@
...
+              `ann` tEmptyHole
+            )
+            `ann` tEmptyHole
+        pure (e0, [(0, e0), (4, e4), (5, e5)])
+   in do
+        for_ @[] expect $ \(s, ex) -> do
+          print s
+          print ex
+          a <- evalFullTestAvoidShadowing maxID mempty mempty s Syn e
+          print a
```
brprice, line 115: I should remove these `print` lines (were a debugging aid that got left in)

PR #712, opened 2022-10-10
feat: logs have trace ids

brprice:
Currently this is somewhat pointless, since we don't actually log anything other than from the db. However, this is set up so that when we do, we will automatically attach the correct trace id.


dhess:
Closing, though I'll keep the branch around for future reference.

Issue #44, opened 2021-07-29
"Pushing down" let bindings


dhess:
Our evaluator should not eliminate `let`s until they're "pushed down" to their use sites. This would permit more flexible interactive evaluation strategies in eval mode (see discussion in https://github.com/hackworthltd/vonnegut/discussions/638). It might also potentially simplify the most complicated bit of the current evaluator &mdash; see https://github.com/hackworthltd/vonnegut/pull/768#discussion_r678488543.

In order to preserve the "small local changes" spirit of our reduction rules in the current eval mode, we should probably implement this as @brprice suggested (and I documented in https://github.com/hackworthltd/vonnegut/discussions/638), such that `let`s float down towards their occurrences. @georgefst also suggested that we could add an affordance to allow students to click-drag a `let` to each of its use sites before eliminating it, which seems like a good idea, though it should probably be part of a suite of similar interactions, rather than just a special case interaction for eliminating `let`s.

But let's start simple. I propose that we change the current `let` elimination rule so that a `let` can only be eliminated once it's adjacent to the use of the bound variable, and that we add a "push down" rule that moves a `let` one step closer to its use(s), splitting the `let` into multiple equivalent `let`s when more than one child contains a use . If the latter is too annoying in practice, we can try some alternate approaches, such as @georgefst's suggestion, or just a macro step that pushes a `let` all the way down to all of its bound variable's occurrences in one go.

(One special case that occurs to me as I write this: Imagine we have `const x y = x` and we evaluate `const 3 2` giving `let x = 3, y = 2 in x` &mdash; we need to be able recognize and step-eliminate the `let y = 2` in this case.)

See #28 for an example of why this feature would be useful.

dhess:
Will be closed by #736.

brprice:
Closed by #736

Issue #734, opened 2022-10-20
Offer multiple reduction choices per subterm, if applicable


brprice:
There are cases where we could say that the same subterm could reduce in two different ways. E.g. `let x = y in case C ... of C y -> x + y`, where we could either reduce the case-of-known-constructor to `let x = y in let y = ... in x + y`, or rename the binders of the case branch so that we could then inline the `let` via `let x = y in case C ... of C z -> let y = z in x + y`. Since EvalFull is the "primary" evaluation method, we want to do the second (as it will, in a few steps, enable us to remove the outer `let`). However, we possibly want to make a different choice (or offer both) for the interactive evaluator.

However, it is not entirely clear whether this will still be an issue when/if we switch to explicit-substitution style pushing-down-lets, #44 . We should revisit it at that point.

_Originally posted by @brprice in https://github.com/hackworthltd/primer/pull/718#discussion_r994574764_

dhess:
I've marked this as high priority not necessarily because it is actually high priority, but just so that we don't lose track of it once #736 is implemented.

Issue #28, opened 2021-07-23
"Pushing down" let bindings


dhess:
### Discussed in https://github.com/hackworthltd/vonnegut/discussions/638

<div type='discussions-op-text'>

<sup>Originally posted by **dhess** June 30, 2021</sup>
Generally speaking, I really like the reduction rule in our evaluation semantics that converts applications to let bindings, as this rule obviates the need for environments (which are not first class in any programming language I'm aware of). However, in Keybase chat today, I pointed out the following limitation of this rule. Consider the following evaluation step in the application of `map` to a list:

<img width="1132" alt="Screen Shot 2021-06-29 at 11 06 53 AM" src="https://user-images.githubusercontent.com/2438/123880082-d849a380-d939-11eb-9b68-fe9eab93348f.png">

Now let's evaluate the `APP` redex such that we get a `lettype β = Nat in ...` in its place:

<img width="1175" alt="Screen Shot 2021-06-29 at 11 08 12 AM" src="https://user-images.githubusercontent.com/2438/123880150-ff07da00-d939-11eb-9b27-5a79c13bb845.png">

The problem here is that I have to reduce the `lettype β = Nat in ...` for *all* occurrences of `β` in the subtree under the `lettype` before I can perform the function application `(λf. ...) (λy.Zero)`. So if I want to mimic the call-by-name evaluation strategy, I can't. I can effectively only do call-by-value.

(Here I use our slightly odd `lettype` form, but the same goes for `let`s for term bindings.)

I proposed that, instead of the current rule which replaces the application with a single `lettype` at the application location in the tree, we push the `lettype β` down to each occurrence of `β` in the subtree, which would allow the student to explore different evaluation strategies.

@georgefst pointed out that in Harry's original "Evaluation steps in Vonnegut" document in Craft (the Apple URI scheme link to the relevant section of that document is craftdocs://open?blockId=CECD7E89-232B-4F44-9A6F-53DF188B9212&spaceId=8b43f204-aeeb-b8ce-45cc-73a653745299), under the "Making substitution incremental" heading, the major rationale for this reduction rule was to not change too much in the tree at once, making application easier to follow; or, as @georgefst put it, "it keeps the evaluation step small and local." By pushing the `let` or `lettype` down to each occurrence of the bound variable, we would be harming this nice pedagogical affordance.

[...]

PR #736, opened 2022-10-20
feat: push down lets

brprice:
This addresses #44 

Notes for reviewers: I suggest going commit-by-commit, but first have a look at my comments-on-code below, some of which have come detached from the code (and thus won't show up in the commit-by-commit review) due to a rebase.

### DIFF
--- a/primer/primer.cabal
+++ b/primer/primer.cabal
@@ -79,6 +79,7 @@ library
     Primer.Eval.Let
     Primer.Eval.NormalOrder
     Primer.Eval.Prim
+    Primer.Eval.Push
     Primer.Primitives.PrimDef
     Primer.Typecheck.Cxt
     Primer.Typecheck.Kindcheck
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -66,7 +66,7 @@ import Primer.Eval.Redex (
   EvalLog (..),
   MonadEval,
   RunRedexOptions (RunRedexOptions),
-  ViewRedexOptions (ViewRedexOptions, groupedLets),
+  ViewRedexOptions (ViewRedexOptions, aggressiveElision, groupedLets),
   getNonCapturedLocal,
   runRedex,
   runRedexTy,
@@ -124,7 +124,11 @@ findNodeByID i =
 -- We hardcode a permissive set of options for the interactive eval
 -- (i.e. these see more redexes)
 evalOpts :: ViewRedexOptions
-evalOpts = ViewRedexOptions{groupedLets = True}
+evalOpts =
+  ViewRedexOptions
+    { groupedLets = True
+    , aggressiveElision = True
+    }
 
 -- | Return the IDs of nodes which are reducible.
 -- We assume that the expression is well scoped. There are no
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -56,7 +56,7 @@ import Primer.Eval.Detail (
  )
 import Primer.Eval.EvalError (EvalError (..))
 import Primer.Eval.NormalOrder (
-  FMExpr (FMExpr, expr, subst, substTy, ty),
+  FMExpr (FMExpr, expr, ty),
   foldMapExpr,
   singletonCxt,
  )
@@ -117,8 +117,6 @@ findNodeByID i =
     FMExpr
       { expr = \ez d c -> if getID ez == i then Just (c, Left (d, ez)) else Nothing
       , ty = \tz c -> if getID tz == i then Just (c, Right tz) else Nothing
-      , subst = Nothing
-      , substTy = Nothing
       }
 
 -- We hardcode a permissive set of options for the interactive eval
@@ -148,8 +146,6 @@ redexes tydefs globals =
       FMExpr
         { expr = \ez d -> liftMaybeT . runReaderT (getID ez <$ viewRedex evalOpts tydefs globals d (target ez))
         , ty = \tz -> runReader (whenJust (getID tz) <$> viewRedexType evalOpts (target tz))
-        , subst = Nothing
-        , substTy = Nothing
         }
   where
     liftMaybeT :: Monad m' => MaybeT m' a -> ListT m' a
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -66,7 +66,7 @@ import Primer.Eval.Redex (
   EvalLog (..),
   MonadEval,
   RunRedexOptions (RunRedexOptions),
-  ViewRedexOptions (ViewRedexOptions),
+  ViewRedexOptions (ViewRedexOptions, groupedLets),
   getNonCapturedLocal,
   runRedex,
   runRedexTy,
@@ -124,7 +124,7 @@ findNodeByID i =
 -- We hardcode a permissive set of options for the interactive eval
 -- (i.e. these see more redexes)
 evalOpts :: ViewRedexOptions
-evalOpts = ViewRedexOptions{}
+evalOpts = ViewRedexOptions{groupedLets = True}
 
 -- | Return the IDs of nodes which are reducible.
 -- We assume that the expression is well scoped. There are no
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -65,7 +65,7 @@ import Primer.Eval.Redex (
   Dir (..),
   EvalLog (..),
   MonadEval,
-  RunRedexOptions (RunRedexOptions),
+  RunRedexOptions (RunRedexOptions, pushAndElide),
   ViewRedexOptions (ViewRedexOptions, aggressiveElision, groupedLets),
   getNonCapturedLocal,
   runRedex,
@@ -159,7 +159,15 @@ redexes tydefs globals =
 
 -- We hardcode a particular set of reduction options for the interactive evaluator
 reductionOpts :: RunRedexOptions
-reductionOpts = RunRedexOptions{}
+reductionOpts =
+  RunRedexOptions
+    { -- For intearctive use, we think combining these two steps is too confusing.
+      -- The choice of hardcoding this makes this feature slightly harder to test,
+      -- see https://github.com/hackworthltd/primer/pull/736#discussion_r1293290757
+      -- for some tests that we would like to have added, if it were simple to test
+      -- a single step of pushAndElide.
+      pushAndElide = False
+    }
 
 -- | Given a context of local and global variables and an expression, try to reduce that expression.
 -- Expects that the expression is redex and will throw an error if not.
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -5,6 +5,8 @@ module Primer.Eval (
   -- The public API of this module
   step,
   redexes,
+  RunRedexOptions (..),
+  ViewRedexOptions (..),
   EvalLog (..),
   EvalError (..),
   EvalDetail (..),
@@ -63,6 +65,8 @@ import Primer.Eval.Redex (
   Dir (..),
   EvalLog (..),
   MonadEval,
+  RunRedexOptions (RunRedexOptions),
+  ViewRedexOptions (ViewRedexOptions),
   getNonCapturedLocal,
   runRedex,
   runRedexTy,
@@ -117,6 +121,11 @@ findNodeByID i =
       , substTy = Nothing
       }
 
+-- We hardcode a permissive set of options for the interactive eval
+-- (i.e. these see more redexes)
+evalOpts :: ViewRedexOptions
+evalOpts = ViewRedexOptions{}
+
 -- | Return the IDs of nodes which are reducible.
 -- We assume that the expression is well scoped. There are no
 -- guarantees about whether we will claim that an ill-sorted variable
@@ -133,8 +142,8 @@ redexes tydefs globals =
   (ListT.toList .)
     . foldMapExpr
       FMExpr
-        { expr = \ez d -> liftMaybeT . runReaderT (getID ez <$ viewRedex tydefs globals d (target ez))
-        , ty = \tz -> runReader (whenJust (getID tz) <$> viewRedexType (target tz))
+        { expr = \ez d -> liftMaybeT . runReaderT (getID ez <$ viewRedex evalOpts tydefs globals d (target ez))
+        , ty = \tz -> runReader (whenJust (getID tz) <$> viewRedexType evalOpts (target tz))
         , subst = Nothing
         , substTy = Nothing
         }
@@ -143,6 +152,10 @@ redexes tydefs globals =
     liftMaybeT m = ListT $ fmap (,mempty) <$> runMaybeT m
     -- whenJust :: Alternative f => a -> Maybe b -> f a
     whenJust = maybe empty . const . pure
+
+-- We hardcode a particular set of reduction options for the interactive evaluator
+reductionOpts :: RunRedexOptions
+reductionOpts = RunRedexOptions{}
 
 -- | Given a context of local and global variables and an expression, try to reduce that expression.
 -- Expects that the expression is redex and will throw an error if not.
@@ -156,8 +169,8 @@ tryReduceExpr ::
   Expr ->
   m (Expr, EvalDetail)
 tryReduceExpr tydefs globals cxt dir expr =
-  runMaybeT (flip runReaderT cxt $ viewRedex tydefs globals dir expr) >>= \case
-    Just r -> runRedex r
+  runMaybeT (flip runReaderT cxt $ viewRedex evalOpts tydefs globals dir expr) >>= \case
+    Just r -> runRedex reductionOpts r
     _ -> throwError NotRedex
 
 tryReduceType ::
@@ -169,6 +182,6 @@ tryReduceType ::
   Type ->
   m (Type, EvalDetail)
 tryReduceType _globals cxt =
-  flip runReader cxt . viewRedexType <&> \case
-    Just r -> runRedexTy r
+  flip runReader cxt . viewRedexType evalOpts <&> \case
+    Just r -> runRedexTy reductionOpts r
     _ -> throwError NotRedex
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -22,7 +22,7 @@ module Primer.Eval (
   -- Only exported for testing
   Cxt (Cxt),
   singletonCxt,
-  getNonCapturedLocal,
+  lookupEnclosingLet,
   tryReduceExpr,
   tryReduceType,
   findNodeByID,
@@ -67,7 +67,7 @@ import Primer.Eval.Redex (
   MonadEval,
   RunRedexOptions (RunRedexOptions, pushAndElide),
   ViewRedexOptions (ViewRedexOptions, aggressiveElision, groupedLets),
-  getNonCapturedLocal,
+  lookupEnclosingLet,
   runRedex,
   runRedexTy,
   viewRedex,
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -107,8 +107,8 @@ step tydefs globals expr d i = runExceptT $ do
       pure (expr', detail)
 
 -- | Search for the given node by its ID.
--- Collect all local bindings in scope and return them
--- (with their local definition, if applicable)
+-- Collect all immediately-surrounding let bindings and return them
+-- (these are the ones we may push into this node)
 -- along with the focused node.
 -- Returns Nothing if the node is a binding, (note that no reduction rules can apply there).
 findNodeByID :: ID -> Dir -> Expr -> Maybe (Cxt, Either (Dir, ExprZ) TypeZ)
--- a/primer/src/Primer/Eval.hs
+++ b/primer/src/Primer/Eval.hs
@@ -16,6 +16,7 @@ module Primer.Eval (
   GlobalVarInlineDetail (..),
   LetRemovalDetail (..),
   ApplyPrimFunDetail (..),
+  PushLetDetail (..),
   -- Only exported for testing
   Cxt (Cxt),
   singletonCxt,
@@ -49,6 +50,7 @@ import Primer.Eval.Detail (
   GlobalVarInlineDetail (..),
   LetRemovalDetail (..),
   LocalVarInlineDetail (..),
+  PushLetDetail (..),
  )
 import Primer.Eval.EvalError (EvalError (..))
 import Primer.Eval.NormalOrder (
--- a/primer/src/Primer/Eval/NormalOrder.hs
+++ b/primer/src/Primer/Eval/NormalOrder.hs
@@ -25,8 +25,6 @@ import Primer.Core (
     LetType,
     Letrec
   ),
-  TyVarName,
-  Type,
   Type' (
     TLet
   ),
@@ -115,11 +113,10 @@ foldMapExpr extract topDir = go mempty . (topDir,) . focus
     go :: Cxt -> (Dir, ExprZ) -> f a
     go lets dez@(d, ez) =
       extract.expr ez d lets
-        <|> case (extract.subst, viewLet dez) of
-          (Just goSubst, Just (ViewLet{bindingVL, bodyVL = (d', b)})) -> goSubst bindingVL b d' $ cxtAddLet bindingVL lets
+        <|> case viewLet dez of
           -- Prefer to compute inside the body of a let, but otherwise compute in the binding
           -- NB: we never push lets into lets, so the Cxt is reset for non-body children
-          (Nothing, Just (ViewLet{bindingVL, bodyVL, typeChildrenVL, termChildrenVL})) ->
+          Just (ViewLet{bindingVL, bodyVL, typeChildrenVL, termChildrenVL}) ->
             msum $
               go (cxtAddLet bindingVL lets) bodyVL
                 : map (goType mempty) typeChildrenVL
@@ -134,10 +131,8 @@ foldMapExpr extract topDir = go mempty . (topDir,) . focus
     goType :: Cxt -> TypeZ -> f a
     goType lets tz =
       extract.ty tz lets
-        <|> case (extract.substTy, target tz) of
-          (Just goSubstTy, TLet _ a t _body)
-            | [_, bz] <- typeChildren tz -> goSubstTy a t bz lets
-          (Nothing, TLet _ a t _body)
+        <|> case target tz of
+          TLet _ a t _body
             -- Prefer to compute inside the body of a let, but otherwise compute in the binding
             | [tz', bz] <- typeChildren tz -> goType (cxtAddLet (LetTyBind $ LetTypeBind a t) lets) bz <|> goType mempty tz'
           _ -> msum $ map (goType mempty) $ typeChildren tz
@@ -145,8 +140,6 @@ foldMapExpr extract topDir = go mempty . (topDir,) . focus
 data FMExpr m = FMExpr
   { expr :: ExprZ -> Dir -> Cxt -> m
   , ty :: TypeZ -> Cxt -> m
-  , subst :: Maybe (LetBinding -> ExprZ {- The body of the let-} -> Dir -> Cxt -> m)
-  , substTy :: Maybe (TyVarName -> Type -> TypeZ -> Cxt -> m)
   }
 
 focusType' :: MonadPlus m => ExprZ -> m TypeZ
@@ -167,8 +160,6 @@ findRedex opts tydefs globals =
     ( FMExpr
         { expr = \ez d -> runReaderT (RExpr ez <$> viewRedex opts tydefs globals d (target ez))
         , ty = \tz -> hoistMaybe . runReader (RType tz <<$>> viewRedexType opts (target tz))
-        , subst = Nothing
-        , substTy = Nothing
         }
     )
 
--- a/primer/src/Primer/Eval/NormalOrder.hs
+++ b/primer/src/Primer/Eval/NormalOrder.hs
@@ -17,7 +17,6 @@ import Control.Monad.Trans.Accum (
   Accum,
   add,
   execAccum,
-  look,
  )
 import Control.Monad.Trans.Maybe (MaybeT)
 import Data.Map qualified as M
@@ -32,13 +31,11 @@ import Primer.Core (
     LetType,
     Letrec
   ),
-  HasID,
   TyVarName,
   Type,
   Type' (
     TLet
   ),
-  getID,
  )
 import Primer.Def (
   DefMap,
@@ -211,16 +208,14 @@ exprChildren (d, ez) =
 typeChildren :: TypeZ -> [TypeZ]
 typeChildren = children'
 
-addBinds :: HasID i => i -> [Either Name LetBinding] -> Accum Cxt ()
-addBinds i' bs = do
-  let i = getID i'
-  cxt <- look
+addBinds :: i -> [Either Name LetBinding] -> Accum Cxt ()
+addBinds _ bs = do
   add $
     Cxt $
       M.fromList $
         bs <&> \case
-          Left n -> (n, (Nothing, i, cxt))
-          Right l -> (letBindingName l, (Just l, i, cxt))
+          Left n -> (n, Nothing)
+          Right l -> (letBindingName l, Just l)
 
-singletonCxt :: HasID i => i -> LetBinding -> Cxt
+singletonCxt :: i -> LetBinding -> Cxt
 singletonCxt i l = addBinds i [Right l] `execAccum` mempty
--- a/primer/src/Primer/Eval/NormalOrder.hs
+++ b/primer/src/Primer/Eval/NormalOrder.hs
@@ -55,6 +55,7 @@ import Primer.Eval.Redex (
   EvalLog,
   Redex,
   RedexType,
+  ViewRedexOptions,
   viewRedex,
   viewRedexType,
  )
@@ -176,16 +177,17 @@ hoistAccum = Foreword.hoistAccum generalize
 findRedex ::
   forall l m.
   (MonadLog (WithSeverity l) m, ConvertLogMessage EvalLog l) =>
+  ViewRedexOptions ->
   TypeDefMap ->
   DefMap ->
   Dir ->
   Expr ->
   MaybeT m RedexWithContext
-findRedex tydefs globals =
+findRedex opts tydefs globals =
   foldMapExpr
     ( FMExpr
-        { expr = \ez d -> runReaderT (RExpr ez <$> viewRedex tydefs globals d (target ez))
-        , ty = \tz -> hoistMaybe . runReader (RType tz <<$>> viewRedexType (target tz))
+        { expr = \ez d -> runReaderT (RExpr ez <$> viewRedex opts tydefs globals d (target ez))
+        , ty = \tz -> hoistMaybe . runReader (RType tz <<$>> viewRedexType opts (target tz))
         , subst = Nothing
         , substTy = Nothing
         }
--- a/primer/src/Primer/Eval/NormalOrder.hs
+++ b/primer/src/Primer/Eval/NormalOrder.hs
@@ -13,13 +13,7 @@ module Primer.Eval.NormalOrder (
 import Foreword hiding (hoistAccum)
 
 import Control.Monad.Log (MonadLog, WithSeverity)
-import Control.Monad.Trans.Accum (
-  Accum,
-  add,
-  execAccum,
- )
 import Control.Monad.Trans.Maybe (MaybeT)
-import Data.Map qualified as M
 import Primer.Core (
   Expr,
   Expr' (
@@ -52,7 +46,6 @@ import Primer.Eval.Redex (
   viewRedexType,
  )
 import Primer.Log (ConvertLogMessage)
-import Primer.Name (Name)
 import Primer.TypeDef (
   TypeDefMap,
  )
@@ -65,7 +58,6 @@ import Primer.Zipper (
   down,
   focus,
   focusType,
-  letBindingName,
   right,
   target,
  )
@@ -208,14 +200,5 @@ exprChildren (d, ez) =
 typeChildren :: TypeZ -> [TypeZ]
 typeChildren = children'
 
-addBinds :: i -> [Either Name LetBinding] -> Accum Cxt ()
-addBinds _ bs = do
-  add $
-    Cxt $
-      M.fromList $
-        bs <&> \case
-          Left n -> (n, Nothing)
-          Right l -> (letBindingName l, Just l)
-
-singletonCxt :: i -> LetBinding -> Cxt
-singletonCxt i l = addBinds i [Right l] `execAccum` mempty
+singletonCxt :: LetBinding -> Cxt
+singletonCxt l = Cxt [l]
--- a/primer/src/Primer/Eval/NormalOrder.hs
+++ b/primer/src/Primer/Eval/NormalOrder.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE DisambiguateRecordFields #-}
-{-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE OverloadedRecordDot #-}
 
 module Primer.Eval.NormalOrder (
@@ -27,16 +26,6 @@ import Control.Monad.Trans.Accum (
  )
 import Control.Monad.Trans.Maybe (MaybeT)
 import Data.Map qualified as M
-import Data.Set qualified as S
-import Data.Set.Optics (setOf)
-import Optics (
-  elemOf,
-  folded,
-  getting,
-  to,
-  (%),
-  _2,
- )
 import Primer.Core (
   Expr,
   Expr' (
@@ -55,11 +44,7 @@ import Primer.Core (
   Type' (
     TLet
   ),
-  bindName,
   getID,
- )
-import Primer.Core.Utils (
-  _freeVarsTy,
  )
 import Primer.Def (
   DefMap,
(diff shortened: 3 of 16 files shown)


primer/primer.cabal
```diff
@@ -79,6 +79,7 @@ library
     Primer.Eval.Let
     Primer.Eval.NormalOrder
     Primer.Eval.Prim
+    Primer.Eval.Push
```
brprice, line 82: A note on the commit message of 'feat: evaluate lets by "pushing down"':  there is a note here about "document the future simplifications" -- I'll reword this later once I have actually done some such

primer/test/Tests/Eval.hs
```diff
@@ -1163,29 +1131,35 @@ unit_redexes_let_capture =
...
+unit_redexes_lettype_capture = do
+  -- We can push the letType down once
+  -- TODO: we don't really want the "forall y" to be a redex -- it is only a
+  -- rename, and not blocking anything yet. It would be preferable to only
+  -- consider renaming binders when they are immediately under the corresponding
+  -- @let@ (actually, immediately under a group of @let@s containing the
+  -- corresponding one).
+  redexesOf (letType "x" (tvar "y") (emptyHole `ann` tforall "y" KType (tvar "x"))) <@?=> Set.fromList [0, 4]
+  -- But now we should rename the forall and not push the tlet further
+  redexesOf (emptyHole `ann` tlet "x" (tvar "y") (tforall "y" KType (tvar "x"))) <@?=> Set.singleton 4
```
brprice: This TODO is fixed in a later commit in this PR

primer/src/Primer/Eval/Redex.hs
```diff
@@ -224,7 +232,7 @@ data Redex
       , varID :: ID
       -- ^ Where was the occurrence (used for details)
       }
```
brprice, line 235: Maybe I should revisit this definition, since recording these details seems slightly silly now they are all local. Similarly for `InlineLetRec` and `InlineLetInType`

primer/test/Tests/Eval.hs
```diff
@@ -936,7 +936,7 @@ unit_findNodeByID_capture_type =
             , Just (1, LetTyBind _) <- lookupCaptured "x" locals
             , Just (5, LetTyBind _) <- lookupCaptured "z" locals ->
                 pure ()
-          Just (_, Right _) -> assertFailure "Expected lettype binding of 'x' and the tlet binding of 'z' to be reported as captured-if-inlined"
+          Just (_, Right _) -> assertFailure "Expected lettype binding of 'x' and the tlet binding of 'z' to be reported as captured-if-inlined" -- TODO: can probably remove all the "captured-if-inlined" stuff as don't do inlining like that now
```
brprice: This is referring to one of the simplifications that I should do & document

primer/src/Primer/Eval/NormalOrder.hs
```diff
@@ -92,29 +85,34 @@ data RedexWithContext
...
-    tz = maybe [] ((: []) . pure) $ focusType ez
+    tz = maybe [] ((: []) . (Nothing,)) $ focusType ez
 
 -- | This is similar to 'foldMap', with a few differences:
 -- - 'Expr' is not foldable
 -- - We target every subexpression and also every (nested) subtype (e.g. in an annotation)
 -- - We keep track of context and directionality (necessitating an extra 'Dir' argument, for "what directional context this whole expression is in")
+-- TODO: the "context" is just of the immediately-enclosing lets, which are the only ones that may "cross a binder"
 -- - We handle @let@s specially, since we need to handle them differently when finding the normal order redex.
+-- TODO: do we still "handle lets specially", i.e. do we ever set subst = Just ...?
```
brprice, line 115: This is some of the simplifications I alluded to earlier. I plan to do these simple ones shortly

primer/src/Primer/Eval/Redex.hs
```diff
@@ -611,6 +612,7 @@ viewCaseRedex tydefs = \case
     formCaseRedex con argTys params args binders rhs (orig, scrut, conID) =
       CaseRedex{con, args, argTys, params, binders, rhs, orig, scrutID = getID scrut, conID}
 
+-- TODO: don't need to record nearly so much anymore -- a list of directly-enclosing let bindings would be good enough
```
brprice, line 615: (ditto about simplifications)

primer/src/Primer/Eval/Redex.hs
```diff
@@ -619,6 +621,9 @@ newtype Cxt = Cxt (M.Map Name (Maybe LetBinding, ID, Cxt))
   -- and want later 'add's to overwrite earlier (more-global) context entries
   deriving (Semigroup, Monoid) via Dual (M.Map Name (Maybe LetBinding, ID, Cxt))
 
+cxtAddLet :: LetBinding -> Cxt -> Cxt
+cxtAddLet l (Cxt c) = Cxt $ M.insert (letBindingName l) (Just l, 0, mempty) c -- TODO: the 0, mempty are LIES, but we never care about these positions...
```
brprice, line 625: This is a temporary hack, and should clear up when I do the simplifications

primer/test/Tests/Eval.hs
```diff
@@ -1459,7 +1437,10 @@ tasty_type_preservation =
 -- | Reductions do not interfere with each other
 -- if @i,j ∈ redexes e@  (and @i /= j@), and @e@ reduces to @e'@ via redex @i@
 -- then @j ∈ redexes e'@,
--- unless @j@ no longer exists in @e'@ or @j@ was a rename-binding which is no longer required
+-- unless one of the following is true
+-- - @j@ no longer exists in @e'@
+-- - @j@ was a rename-binding which is no longer required because @i@ was eliding a let
+-- - @i@ was a PushLet, which can intersperse itself inside the parts of @j@'s redex, blocking it
 tasty_redex_independent :: Property
 tasty_redex_independent =
```
brprice, line 1450: I do wonder if this test is getting a bit silly...

primer/src/Primer/Eval/Redex.hs
```diff
@@ -632,6 +632,22 @@ lookupTy n c = case lookup (unLocalName n) c of
...
+--    (let a = x in Λb. t : let a=x in ∀b. T) @y
+-- one step in type and one in term would give
+--    (Λb. let a = x in t : ∀b. let a=x in T) @y
+-- which would then reduce the top application:
+--    (let b = y in let a = x in t : let b = y in let a=x in T) @y
+-- rather than forcing us to push the substitution through the annotation first.
+-- This way would be more efficient (with grouped lets): we only have to push one large
+-- substitution through T (and t) once, rather than two small ones.
+--
+-- IDEAS: either have an annotation "why stuck", or just say "do let-push in each (ty&tm) child first, before anything else)
```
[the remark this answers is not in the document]
georgefst: I think I might need to wait until you've written this up fully in order to fully understand the trade-off here.

primer/test/Tests/EvalFull.hs
```diff
@@ -1364,6 +1364,12 @@ unit_prim_partial_map =
             `ann` (tcon tList `tapp` tcon tChar)
             <*> pure (M.singleton mapName mapDef)
    in do
+        -- TODO/REVIEW:
+        -- note that ww/map is somewhat pointless when push down lets -- maybe we should push lets inside letrecs (maybe just inside type?)?
+        --     nb, get something like @let α=Char, β=Char, f=toUpper in (letrec go : List α -> List β; go = λxs.RHS in go)
+        --     and then in two steps (expand @go@, push stack of let+letrec)
+        --        @λxs. let α=Char, β=Char, f=toUpper in (letrec go : List α -> List β; go = λxs.RHS in RHS : List α -> List β)
+        --     we carry around the subst for α,β and f, using α,β inside annotation and f in RHS each time expand the letrec
```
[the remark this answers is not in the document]
georgefst: > should I elaborate?

Yes please! I'm struggling to see how the first expression becomes the second ("and then in two steps").

primer/test/Tests/Eval.hs
georgefst: Seeing as you've already written the code to fix it up, I see no good reason not to keep it around.

It's basically just the last case which makes it hard to follow, because of the explicit enumeration of the `i` cases. But then using a wildcard there would make the test less complete.

primer/src/Primer/Eval/Redex.hs
```diff
@@ -250,6 +250,11 @@ data Redex
     --  nor any free variable of @e@ (to avoid capture)]
     -- [We actually do this rule for a whole sequence of let bindings at once]
     -- [If we push into an annotation, we drop term variables:  let x = e in (t : T)  ~> (let x = e in t) : T]
+    -- [We actually drop all "pointless" variables:
+    --   let x = e in f s  ~>  (let x = e in f) (let x = e in s)    if @x@ is free in @f@, and free in @s@
+    --   let x = e in f s  ~>  f (let x = e in s)                   if @x@ is not free in @f@, but free in @s@
+    --   let x = e in f s  ~>  (let x = e in f) s                   if @x@ is free in @f@, but not free in @s@
+    --   let x = e in f s  ~>  f s                                  if @x@ is not free in @f@, and not free in @s@
```
georgefst, line 255: This does somewhat violate our "eval steps should be as small/simple as possible" maxim. I think it's worth it here though. We'll just need to be careful about how we explain it in the UI.

georgefst (approved):

primer/test/Tests/EvalFull.hs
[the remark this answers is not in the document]
brprice: To elaborate on the above:

Recall that the w/w map is
```
map :: ∀a. ∀b. (a -> b) -> List a -> List b
map = Λa. Λb. λf.
    letrec go :: List a -> List b
           go = λxs. case xs of
                       Nil -> Nil
                       Cons y ys -> Cons (f y) (go ys)
    in go
```
Now consider the application `map Char Char toUpper "abc"`, we get via inlining the global and one BETA step
```
((let a = Char in Λb. λf.
    letrec go :: ...
           go = ...
    in go)
 :: let a = Char in ∀b.(a -> b) -> List a -> List b) Char toUpper "abc"
```
and then we fully reduce the type annotation, before looking at the term (yes this is a bit odd, see my other comment)
```
((let a = Char in Λb. λf.
    letrec go :: ...
           go = ...
    in go)
 :: ∀b.(Char -> b) -> List Char -> List b) Char toUpper "abc"
```
Now one PushLet gives
```
((Λb. let a = Char in λf.
    letrec go :: ...
           go = ...
    in go)
 :: ∀b.(Char -> b) -> List Char -> List b) Char toUpper "abc"
```
which is very similar to the above. Another BETA and reduction of the annotation gives
```
((let b = Char in let a = Char in λf.
    letrec go :: ...
           go = ...
    in go)
 :: (Char -> Char) -> List Char -> List Char) toUpper "abc"
```
Now push the lets again
```
((λf. let b = Char in let a = Char in
    letrec go :: ...
           go = ...
    in go)
 :: (Char -> Char) -> List Char -> List Char) toUpper "abc"
```
And a Beta step
```
((let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
    letrec go :: ...
           go = ...
    in go)
 :: List Char -> List Char) "abc"
```
Now (and this is the first step of what I above mentioned as "in two steps") we expand the `go`
```
((let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
    letrec go :: ...
           go = λxs. ...
    in λxs. ...)
 :: List Char -> List Char) "abc"
```
and push the lets
```
((λxs. let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
    letrec go :: List a -> List b
           go = λxs. ... Cons (f y) (go ys)
    in case ...)
 :: List Char -> List Char) "abc"
```
Note that this will now reduce by Beta-reducing the `λxs`, and we will keep the stack of lets `f,b,a` around because they are used in the annotation and body of the letrec.
Before we changed to pushing-down-lets, we would have inlined `f,b,a`, giving `letrec go :: List Char -> List Char; go = λxs. case xs of Nil -> Nil ; Cons y ys -> Cons (toUpper y) (go ys)`, which is smaller and faster.

<details>
<summary>Some selected steps in the rest of the reduction, eliding some annotations:</summary>
```
let xs = "abc" in let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
    letrec go :: List a -> List b
           go = λxs. ... Cons (f y) (go ys)
    in case ...
```
```
case (let xs = "abc" in xs) of
  Nil -> Nil
  Cons y ys -> let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
               letrec go :: List a -> List b
                      go = λxs. case xs of
                                  Nil -> Nil
                                  Cons y ys -> Cons (f y) (go ys)
               in Cons (f y) (go ys)
```
```
case "abc" of
  Nil -> Nil
  Cons y ys -> let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
               letrec go :: List a -> List b
                      go = λxs. case xs of
                                  Nil -> Nil
                                  Cons y ys -> Cons (f y) (go ys)
               in Cons (f y) (go ys)
```
```
let y='a' in let ys="bc" in
let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
letrec go :: List a -> List b
       go = λxs. case xs of
                    Nil -> Nil
                    Cons y ys -> Cons (f y) (go ys)
in Cons (f y) (go ys)
```
```
Cons (let y='a' in let f = toUpper in f y)
     (let ys="bc" in
      let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
      letrec go :: List a -> List b
             go = λxs. case xs of
                         Nil -> Nil
                         Cons y ys -> Cons (f y) (go ys)
      in go ys)
```
```
Cons 'A'
     ((let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
       letrec go :: List a -> List b
              go = λxs. case xs of
                          Nil -> Nil
                          Cons y ys -> Cons (f y) (go ys)
       in go)
      (let ys="bc" in ys))
```
```
Cons 'A'
  (Cons 'B'
     ((let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
       letrec go :: List a -> List b
              go = λxs. case xs of
                          Nil -> Nil
                          Cons y ys -> Cons (f y) (go ys)
       in go)
      (let ys="c" in ys)))
```
```
Cons 'A'
  (Cons 'B'
    (Cons 'C'
     ((let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
       letrec go :: List a -> List b
              go = λxs. case xs of
                          Nil -> Nil
                          Cons y ys -> Cons (f y) (go ys)
       in go)
      (let ys=Nil in ys))))
```
```
Cons 'A'
  (Cons 'B'
    (Cons 'C'
      (case (let xs = Nil in xs) of
         Nil -> Nil
         Cons y ys -> let f = (toUpper :: Char -> Char) in let b = Char in let a = Char in
                      letrec go :: List a -> List b
                             go = λxs. case xs of
                                         Nil -> Nil
                                         Cons y ys -> Cons (f y) (go ys)
                      in Cons (f y) (go ys))))
```
```
Cons 'A' (Cons 'B' (Cons 'C' Nil))
```
</details>

primer/src/Primer/Eval/Redex.hs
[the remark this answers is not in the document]
brprice: The FR #1097 has been opened.

primer/src/Primer/Eval/Redex.hs
```diff
@@ -610,18 +609,14 @@ viewCaseRedex tydefs = \case
...
--- We record each directly-enclosing let binder, along with its let-bound RHS (wrapped in @Just@ for historical reasons)
--- Invariant: lookup x c == Just (Just l) ==> letBindingName l == x
+-- We record each directly-enclosing let binding
 -- By "directly enclosing" we mean "those which may be pushed into this term"
-newtype Cxt = Cxt (M.Map Name (Maybe LetBinding))
-  -- We want right-biased mappend, as we will use this with 'Accum'
-  -- and want later 'add's to overwrite earlier (more-global) context entries
-  deriving (Semigroup, Monoid) via Dual (M.Map Name (Maybe LetBinding))
+-- NB: we do not care about ordering
+newtype Cxt = Cxt [LetBinding]
```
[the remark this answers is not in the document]
brprice: I don't really see the point -- we don't use it in a set-like way, only really adding single bindings and calculating `cxtToAvoid`. We could probably replace it with a set of names (i.e. bake out the `cxtToAvoid` computation). I suspect the overhead of a Set would not be worth it here.

dhess:
FYI, this is a long-standing feature request and we've forgotten some of the details about why: see here: https://github.com/hackworthltd/primer/issues/28

primer/src/Primer/Eval/Redex.hs
dhess: Per discussion in our 2023-07-25 developer meeting:

* We do this in the full evaluator for performance reasons, to prevent evaluation of a pointless push down
* There is some concern that in step eval, it will not be clear why the `let` isn't pushed down into both subtrees
* Therefore in step eval, pushing down a `let` will push down into both subtrees

This marks a branch in the evaluator's development: up to this point, the full and step evaluators have shared the same implementation. However, we already know that they will need to diverge in the near future so that we can start making optimizations in the full evaluator (its performance is too slow to be practical), so this divergence was inevitable anyway.

@georgefst points out that this may be ... (will return to this comment in a bit)

primer/test/Tests/EvalFull.hs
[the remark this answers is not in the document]
brprice: TODO: I will write a note about this

primer/src/Primer/Eval/Redex.hs
brprice: In a meeting today we decided:
- [ ] I'll document the performance impact
- [ ] We will make it optional, so full eval will push+elide but step eval will simply push (and elide in a separate step)

We also discussed the similar violation where we simultaneously push multiple adjacent `let`s, and for those I will also
- [ ] Document performance impact
- [ ] Document the fact we simultaneously push (it is not super obvious!) -- probably write a note and link from the option definition and also the implementation
- [ ] make it optional, in same way as above

Note that we have two performance-impacting things, and historically the push+elide was implemented first, to fix the perf regression in `unit_8`. These two obviously interact, and I am not actually sure how much of the impact comes from each -- I should probably get 4 perf numbers (4 = 2*2 option settings).

brprice: One interesting unintentional interaction is that non-simultaneous-push is actually more aggressive with eliding redundant lets. In `let a = _ in let b = _ in λc. b` with simultaneous-push we will push both `let`s under the lambda. But with non-simultaneous-push we will consider whether 'let a' is a redex and decide it is not a `PushLet` redex as there is a stack of lets underneath it, but it is an `ElideLet` redex (as we fall through to the next case of `viewRedex`). This is an inconsistency that gives extremely misleading performance numbers.

Now that I understand what is happening, I can obtain sensible measurements.

brprice: Some timing data follows.

I ran `mapEven` for various options and lengths of lists (this is roughly `unit_8`), and also the "worker/wrapper"/`unit_9` variation. This was done with a very simple script, so the numbers are probably somewhat noisy.
(Note that we have 3 options now: the two mentioned above, and "aggressiveElision" which prefers to elide early, rather tan pushing all the way to a leaf before eliding)

For `n=5`, I get 
pushMulti  |  aggressiveElision  |  pushAndElide | map, steps | map, time (s) | map/ww, steps | map/ww, time (s)
-----------|---------------------|---------------|------------|---------------|---------------|-----------------
True | True | True | 450 | 0.129396629000 | 368 | 0.130995374000
False | True | True | 450 | 0.118675216000 | 373 | 0.111896999000
True | False | True | 450 | 0.130649372000 | 368 | 0.131270030000
False | False | True | 450 | 0.116904985000 | 373 | 0.110885103000
True | True | False | 583 | 0.170570243000 | 461 | 0.162431173000
False | True | False | 628 | 0.169462742000 | 518 | 0.157327909000
True | False | False | 687 | 1.926900242000 | 500 | 0.440785217000
False | False | False | 997 | 2.959247222000 | 1465 | 4.398923352000

And for `n=10` I  get
pushMulti  |  aggressiveElision  |  pushAndElide | map, steps | map, time (s) | map/ww, steps | map/ww, time (s)                                                                                                                                                                                                               
-----------|---------------------|---------------|------------|---------------|---------------|-----------------                                                                                                                                                                                                               
True | True | True | 1145 | 0.389490347000 | 923 | 0.357104768000                                                                                                                                                                                                                                                              
False | True | True | 1145 | 0.364744574000 | 933 | 0.320479639000                                                                                                                                                                                                                                                             
True | False | True | 1145 | 0.385016614000 | 923 | 0.358131104000                                                                                                                                                                                                                                                             
False | False | True | 1145 | 0.363208693000 | 933 | 0.316983115000                                                                                                                                                                                                                                                            
True | True | False | 1478 | 0.540102081000 | 1151 | 0.455796149000                                                                                                                                                                                                                                                            
False | True | False | 1588 | 0.566575896000 | 1258 | 0.459367898000                                                                                                                                                                                                                                                           
True | False | False | 1707 | 518.533781665000 | 1250 | 4.326406590000 

(The `False,False,False` configuration` took over 2000 steps)

dhess: So my brief analysis is:

* `map/ww` is between 39% to 85% worse performance with push-down `let`s, in the best case scenario (F/F/T). It is still faster than straightforward `map`, however, except for a few odd cases with `n = 5` where I assume there is just too much noise.
* `map` is between to  14% to 25% worse performance with push-down `let`s, in the best case scenario (also F/F/T).
* The performance hit of push-down `let`s decreases as `n` increases.
* For these benchmarks, at least, enabling `pushAndElide` and disabling the other optimizations (i.e., F/F/T) is the winner.
* For these benchmarks, enabling `pushMulti` appears to always be worse, and when only that feature is enabled, it is several orders of magnitude worse on the `map` benchmark, which is weird. @brprice do you have a handle on why this might be?

This is a difficult call. My intuition is that pushing down `let`s is going to be a good pedagogical feature, for the reasons explained in #28, but assuming that the performance impact seen on this benchmark is indicative, push-down `let`s will come with a not insignificant performance cost, at a time when we're particularly concerned about improving performance.

@brprice would it be feasible to keep around the current behavior as yet another option? Ideally we could keep it around for both full- and step-evaluation, but if it were easier only to keep it around for full eval, that would also be OK, I think, and fairly easily explainable as an optimization that the full evaluator can perform.

dhess:
Just a reminder: @brprice had said in #44 that once we implement push-down `let`s, we should think about https://github.com/hackworthltd/primer/issues/734.

primer/src/Primer/Eval/Redex.hs
georgefst: > The performance hit of push-down `let`s decreases as `n` increases.

I think it might be useful to see a comparison of the old behaviour to the new one with all optimisations, on larger programs. Then we can check whether these continue to converge.

dhess: That's an interesting point. It could be that once we've optimized the full evaluator that, for large enough `n` such that the runtime is still noticeable, the performance impact of push-down `let`s is insignificant.

brprice: Some new numbers (I fixed a silly bug...)

Running the whole testsuite:
      | no PBTs (via hyperfine) | with PBTs (three manual runs, sorted)
old | 482.1 ms ±   8.5 ms | 73.4s, 76.6s, 85.6s
new | 576.6 ms ±  29.9 ms | 85.9s, 106.6s, 116.3s

I'll edit some longer runs in here shortly...

dhess: OK, so to make sure I understand correctly, pretty much all the analysis I did earlier is invalid? :)

brprice: I'm afraid so...

Last batch of numbers: memory usage. This one is very in favor of the new approach

With `n=100`:
|| Total alloc bytes | copied during gc bytes | max residency bytes | total memory in use | productivity
|---|---|---|---|---|---
map, this PR | 251,370,361,304 | 45,984,704,024 | 11,501,960 | 52 MiB | 59.2%
map, orig | 192,034,025,032 | 35,501,492,712 | 400,189,696 |  1112 MiB |  55.6%
map/ww, this PR |206,965,550,480| 41,429,113,248|12,180,304|52 MiB|58.0%
map/ww, orig |168,011,848,872|33,181,823,024|358,839,832|994 MiB| 54.1%

georgefst: > This one is very in favor of the new approach

That's a pretty dramatic difference - any idea why?

brprice: I was expecting some difference in favor of the new approach because we simply have to remember less when reducing. Since our rules are now local, we don't need to carry around a context of all (distantly) enclosing `let`s and which ones can be substituted without causing capture. 

However, I am surprised the difference is this large.

brprice: In our meeting today, we decided to go ahead with this PR regardless of the performance impact. Our reasoning was that 
- we are aiming for pedagogy of eval, rather than performance
- the much lower memory footprint may (perhaps) be more important to overall performance than the speed (for memory-constrained systems)
- the "full evaluator" is a simple wrapper around the interactive/step evaluator -- if we want to improve performance, we should change strategy (eval on a virtual machine; do some compilation; etc), though it is not clear if/how this could be integrated into the step eval as a "jump forward"/"don't show me steps inside that subexpression" (due to (perhaps) wanting to stop before reaching a normal form and needing to reify a result back to a term; also due to (perhaps) wanting to evaluate open terms (c.f. #1055))

brprice: When rebasing this PR, I have found that the above memory impact was actually an orthogonal change, which is now contained in #1110. I will generate some new numbers shortly.

brprice: New numbers, for both before and after this PR. Both are based off the new strictness annotation, #1110:

Memory, generated by
- `unit tests`: `cabal run -O2 primer-test -- --hide-successes --hedgehog-tests 0 +RTS -s` (only for the default set of options, as there are some tests that require such)
- `big 8, n=...`: `cabal run -O2 primer-test -- --hide-successes -p EvalFull.8 +RTS -s`, after editing `EvalFull.unit_8` to change `n` and bump the upper bound from 2000 to 200000 and `EvalFull.evalFullTest` to set the options

| Branch  | Test suite | push Multi | aggr. Elision | push And Elide | alloc heap (B) | copy gc (B) | max res (B) | max slop (B) | total mem (MB) |
|:--------|:-----------|----------:|------------------:|-------------:|-----------:|--------:|--------:|---------:|----------:|
| main    | unit tests | N/A  | N/A  | N/A  | 2,253,549,080 | 81,112,304 | 5,381,576 | 245,432 | 34 | 
| this pr | unit tests | True | True | True | 2,254,933,496 | 85,541,920 | 4,884,232 | 120,912 | 33 | 
| main    | big 8, 10  | N/A  | N/A  | N/A  | 687,433,096 | 14,499,296 | 1,679,576 | 61,224 | 23 | 
| this pr | big 8, 10  | True | True | True | 667,293,888 | 23,986,928 | 1,806,288 | 76,024 | 24 | 
| this pr | big 8, 10  | True | True | False| 951,432,832 | 54,166,384 | 2,183,864 | 75,792 | 24 | 
| this pr | big 8, 10  | True | False| True | 667,293,912 | 24,121,216 | 1,835,000 | 67,728 | 24 | 
| this pr | big 8, 10  | False| True | True | 726,874,408 | 20,025,288 | 1,807,416 | 76,744 | 25 | 
| this pr | big 8, 10  | False| True | False| 1,060,585,984 | 41,553,904 | 1,913,520 | 74,824 | 25 | 
| this pr | big 8, 10  | False| False| True | 723,390,560 | 20,034,168 | 1,806,304 | 69,664 | 25 | 
| main    | big 8, 20  | N/A  | N/A  | N/A  | 2,590,407,608 | 73,241,792 | 2,105,768 | 73,304 | 24 | 
| this pr | big 8, 20  | True | True | True | 2,581,434,568 | 126,270,680 | 2,276,296 | 81,544 | 26 | 
| this pr | big 8, 20  | True | True | False| 3,920,475,240 | 334,034,784 | 3,275,976 | 95,000 | 28 | 
| this pr | big 8, 20  | True | False| True | 2,581,434,584 | 127,493,016 | 2,181,192 | 80,168 | 25 | 
| this pr | big 8, 20  | False| True | True | 2,771,095,368 | 108,475,248 | 2,264,312 | 144,136 | 26 | 
| this pr | big 8, 20  | False| True | False| 4,363,811,480 | 294,687,904 | 3,085,192 | 128,832 | 28 | 
| this pr | big 8, 20  | False| False| True | 2,762,316,576 | 108,085,664 | 2,202,808 | 75,832 | 26 | 
| main    | big 8, 50  | N/A  | N/A  | N/A  | 27,423,073,040 | 2,424,616,608 | 4,649,376 | 496,840 | 32 | 
| this pr | big 8, 50  | True | True | True | 27,014,099,816 | 3,321,942,792 | 5,212,016 | 551,048 | 34 | 
| this pr | big 8, 50  | True | True | False| 40,758,062,920 | 7,319,097,552 | 9,713,800 | 672,720 | 46 | 
| this pr | big 8, 50  | True | False| True | 27,014,128,744 | 3,301,876,928 | 5,222,072 | 552,240 | 37 | 
| this pr | big 8, 50  | False| True | True | 28,017,813,904 | 3,098,616,752 | 5,175,104 | 530,616 | 35 | 
| this pr | big 8, 50  | False| True | False| 44,494,569,304 | 7,371,045,888 | 9,545,024 | 604,440 | 45 | 
| this pr | big 8, 50  | False| False| True | 27,964,518,976 | 3,117,906,312 | 5,248,384 | 546,112 | 36 | 
| main    | big 8, 100 | N/A  | N/A  | N/A  | 195,684,938,608 | 34,688,595,528 | 13,795,496 | 1,632,720 | 60 | 
| this pr | big 8, 100 | True | True | True | 190,533,645,264 | 43,294,432,904 | 15,906,656 | 1,651,328 | 68 | 
| this pr | big 8, 100 | True | True | False| 283,194,326,656 | 89,053,479,520 | 33,821,448 | 2,200,456 | 114 | 
| this pr | big 8, 100 | True | False| True | 190,533,646,264 | 43,274,133,832 | 17,822,504 | 1,760,472 | 70 | 
| this pr | big 8, 100 | False| True | True | 194,584,842,160 | 42,208,027,712 | 17,695,144 | 1,715,800 | 70 | 
| this pr | big 8, 100 | False| True | False| 306,139,099,160 | 93,087,751,416 | 32,043,520 | 1,914,912 | 111 | 
| this pr | big 8, 100 | False| False| True | 194,292,871,256 | 42,235,823,416 | 17,311,496 | 1,707,968 | 70 | 


Time, via `hyperfine "$(cabal list-bin -O2 primer-test) --..."` where the time is short enough to get sensible results, but just one run for large `n`

| Branch  | Test suite | push Multi | aggr. Elision | push And Elide | Mean | Min | Max |
|:--------|:-----------|----------:|------------------:|-------------:|-----:|----:|----:|
| main    | unit tests | N/A   | N/A   | N/A   | 447.7 ± 9.1 ms | 425.8 ms | 458.2 ms |
| this pr | unit tests | True  | True  | True  | 470.4 ± 13.6 ms | 445.6 ms | 485.4 ms |
| main    | big 8, 10  | N/A   | N/A   | N/A   | 276.5 ± 4.2 ms | 273.9 ms | 284.3 ms |
| this pr | big 8, 10  | True  | True  | True  | 284.6 ± 0.7 ms | 284.1 ms | 286.6 ms |
| this pr | big 8, 10  | True  | True  | False | 408.7 ± 5.3 ms | 404.1 ms | 415.8 ms |
| this pr | big 8, 10  | True  | False | True  | 286.4 ± 3.1 ms | 284.3 ms | 294.8 ms |
| this pr | big 8, 10  | False | True  | True  | 305.8 ± 3.1 ms | 304.1 ms | 314.1 ms |
| this pr | big 8, 10  | False | True  | False | 439.7 ± 7.2 ms | 434.0 ms | 454.8 ms |
| this pr | big 8, 10  | False | False | True  | 298.4 ± 5.1 ms | 294.0 ms | 304.3 ms |
| main    | big 8, 20  | N/A   | N/A   | N/A   | 953.4 ± 9.6 ms | 934.2 ms | 966.8 ms |
| this pr | big 8, 20  | True  | True  | True  |  1.012 ± 0.007 s | 1.004 s | 1.025 s |
| this pr | big 8, 20  | True  | True  | False |  1.626 ± 0.010 s | 1.605 s | 1.635 s |
| this pr | big 8, 20  | True  | False | True  |  1.026 ± 0.006 s | 1.016 s | 1.037 s |
| this pr | big 8, 20  | False | True  | True  |  1.063 ± 0.008 s | 1.054 s | 1.075 s |
| this pr | big 8, 20  | False | True  | False |  1.744 ± 0.009 s | 1.725 s | 1.755 s |
| this pr | big 8, 20  | False | False | True  |  1.055 ± 0.007 s | 1.044 s | 1.065 s |
| main    | big 8, 50  | N/A   | N/A   | N/A   | 10.858 ± 0.062 s | 10.785 s | 10.985 s |
| this pr | big 8, 50  | True  | True  | True  | 11.320 ± 0.053 s | 11.238 s | 11.405 s |
| this pr | big 8, 50  | True  | True  | False | 18.695 ± 0.129 s | 18.499 s | 18.949 s |
| this pr | big 8, 50  | True  | False | True  | 11.332 ± 0.075 s | 11.255 s | 11.475 s |
| this pr | big 8, 50  | False | True  | True  | 11.602 ± 0.052 s | 11.495 s | 11.685 s |
| this pr | big 8, 50  | False | True  | False | 20.096 ± 0.119 s | 19.907 s | 20.228 s |
| this pr | big 8, 50  | False | False | True  | 11.474 ± 0.078 s | 11.366 s | 11.568 s |
| main    | big 8, 100 | N/A  | N/A  | N/A  | 102.97 s (one run) | N/A | N/A |
| this pr | big 8, 100 | True  | True  | True  | 97.34s (one run) | N/A | N/A |
| this pr | big 8, 100 | True  | True  | False | 160.41s (one run) | N/A | N/A |
| this pr | big 8, 100 | True  | False | True  | 97.33s (one run) | N/A | N/A |
| this pr | big 8, 100 | False | True  | True  | 100.89s (one run) | N/A | N/A |
| this pr | big 8, 100 | False | True  | False | 170.40s (one run) | N/A | N/A |
| this pr | big 8, 100 | False | False | True  | 99.44s (one run) | N/A | N/A |

dhess: Yay! So unless I'm missing something, it looks like with the strictness annotation fix in #1110, and with all 3 push-down `let` optimizations enabled, this PR is now more or less a wash for both wall-clock performance and memory allocations, at least for the benchmarks we've considered here. Is that correct?

brprice: Yeah, more-or-less. Which agrees with my initial assessment (phew!)

dhess: That's great news!

primer/src/Primer/Eval/Redex.hs
brprice: Actually, on second thought I don't think we should. These definitions are all internal, and the implementation of `runRedex{,Ty}` is nicer if we do not need to pattern match to find these identifiers.

The reason we need these IDs is to expose them in the "details" over the API, where I think we should still have them for the benefit of dumb frontends.

primer/src/Primer/Eval/Redex.hs
```diff
@@ -181,7 +181,8 @@ import Primer.Zipper.Type (
 
 {- HLINT ignore ViewRedexOptions "Use newtype instead of data" -}
 data ViewRedexOptions = ViewRedexOptions
-  { pushMulti :: Bool
+  { pushMulti :: Bool -- TODO/REVIEW: rename this: groupedLets?
```
brprice: One last bikeshedding thought: I'm not particularly happy calling this `pushMulti`, since it will also affect elision -- if we are aggressively eliding, then with `pushMulti == True` we will elide everything from a group rather than one-at-a-time.

Any better ideas for names?

primer/src/Primer/Eval/Redex.hs
```diff
@@ -794,6 +795,7 @@ viewLets e = do
       LetType m a ty b -> Just ((m, LetTyBind (LetTypeBind a ty)), b)
       _ -> Nothing
 
+-- TODO: rename?
 unviewLets :: [(ExprMeta, LetBinding)] -> Expr -> Expr
```
brprice: Continuing the last bikeshedding thought: I'm not particularly happy with this name. Currently it is ok as it is the inverse of `viewLets`, but later we tweak `viewLets` so it is not. Elsewhere we have `{un,fold}TApp` as a similar view (though that takes a monadic argument to fill in metas). Perhaps keep `viewLets` (as it will be tweaked later` and change this to `foldLetBindings`?

Any better ideas?

primer/src/Primer/Eval/Redex.hs
```diff
@@ -818,6 +820,7 @@ viewLetsTy ty = do
       Nothing -> ([], t)
       Just (l, t') -> first (l :) $ viewLets' t'
 
+-- TODO: rename?
 unviewLetsTy :: [(TypeMeta, LetTypeBinding)] -> Type -> Type
```
brprice: Essentially the same naming concerns apply here as to `unviewLets`

primer/src/Primer/Eval.hs
```diff
@@ -159,7 +159,7 @@ redexes tydefs globals =
 
 -- We hardcode a particular set of reduction options for the interactive evaluator
 reductionOpts :: RunRedexOptions
-reductionOpts = RunRedexOptions{}
+reductionOpts = RunRedexOptions{pushAndElide = False}
```
brprice: In https://github.com/hackworthltd/primer/pull/736#discussion_r1273386023 we decided that the interactive eval should not do the pushAndElide bigger step. This has been implemented by hardcoding it (rather than making it an option), but this choice makes it slightly harder to test: previously the single-step eval testsuite `tests/Tests/Eval.hs` did `pushAndElide` and had the following tests of this functionality which now don't make sense
```
-- we should not elide @let x = ...@ here, as it is required for @let y@ which is not elided
unit_tryReduce_push_elide_1 :: Assertion
unit_tryReduce_push_elide_1 = do
  let (expr, i) =
        create $ let_ "x" (con0 cTrue) $ let_ "y" (lvar "x") $ lam "z" $ lvar "y"
  result <- runTryReduce tydefs mempty mempty (expr, i)
  case result of
    Right (expr', _) -> expr' ~= create' (lam "z" $ let_ "x" (con0 cTrue) $ let_ "y" (lvar "x") $ lvar "y")
    _ -> assertFailure $ show result

-- we should elide the outer @let x = ...@ in one branch, and the inner in the other
unit_tryReduce_push_elide_2 :: Assertion
unit_tryReduce_push_elide_2 = do
  let (expr, i) =
        create $ let_ "x" (con0 cTrue) $ let_ "y" (lvar "x") $ let_ "x" (con0 cFalse) $ app (lvar "x") (lvar "y")
  result <- runTryReduce tydefs mempty mempty (expr, i)
  case result of
    Right (expr', _) ->
      expr'
        ~= create'
          ( app
              (let_ "x" (con0 cFalse) $ lvar "x")
              (let_ "x" (con0 cTrue) $ let_ "y" (lvar "x") $ lvar "y")
          )
    _ -> assertFailure $ show result

-- we should not elide the outer @let x = ...@ here
unit_tryReduce_push_elide_3 :: Assertion
unit_tryReduce_push_elide_3 = do
  let (expr, i) =
        create $ let_ "x" (con0 cTrue) $ let_ "x" (lvar "x") $ lam "z" $ lvar "x"
  result <- runTryReduce tydefs mempty mempty (expr, i)
  case result of
    Right (expr', _) -> expr' ~= create' (lam "z" $ let_ "x" (con0 cTrue) $ let_ "x" (lvar "x") $ lvar "x")
    _ -> assertFailure $ show result
```
Additionally, the `unit_tryReduce_lettype_self_capture` needs changing if the step eval tests do push-and-elide.

primer/test/Tests/EvalFull.hs
```diff
@@ -1464,6 +1461,17 @@ evalFullTest id_ tydefs globals n d e = do
...
+evalFullTestExactSteps id_ tydefs globals n d e = do
+  s <- evalFullTest id_ tydefs globals (n - 1) d e
+  case s of
+    Right s' -> assertFailure $ "Unexpectedly reached normal form: " <> show s'
+    Left _ -> do
+      t <- evalFullTest id_ tydefs globals n d e
+      case t of
+        Left t' -> assertFailure $ "Unexpected timeout: " <> show t'
+        Right t' -> pure t'
+
```
georgefst, line 1626: Typo in commit message: "teast".

primer/test/Tests/Eval.hs
georgefst: Is this done now?

georgefst: Same goes for the other three discussion threads mentioning "simplifications".

primer/src/Primer/Eval/Redex.hs
georgefst: I think `groupedLets` is better.

primer/src/Primer/Eval.hs
dhess: Can you leave a comment in the source where the hard-code is written, pointing back to the parent comment? (i.e., https://github.com/hackworthltd/primer/pull/736#discussion_r1293290757)

primer/src/Primer/Eval/Redex.hs
georgefst: `foldLetBindings` sounds better. I don't have any other ideas.

georgefst (approved):
Looks good!

I'm very relieved that performance is now okay (https://github.com/hackworthltd/primer/pull/736#discussion_r1291239783). It would have been a shame to have to keep the old version around, given how much is simplified by avoiding non-local substitutions.

primer/test/Tests/Eval.hs
brprice: Yes, in the later commits (this one is in "change Eval.Cxt to match push-down lets")

PR #760, opened 2022-11-03
chore: Migrate to `mtl-2.3.1`

dhess:
Currently blocked on (at least):

* https://github.com/haskell-servant/servant/issues/1636


dhess:
Via #1113, we're now on `mtl-2.3.1` (finally).

Issue #402, opened 2022-04-19
Editable typedefs: type parameters


georgefst:
From "not in spec" section of https://github.com/hackworthltd/primer/issues/267:

> Another case to handle is a student adding a parameter to a type: data T x = A -> data T x y = A.
>
> Note that any fresh parameter is, of course, initially unused.
>
> We may want an easy way to allow making a type more or less polymorphic. e.g. converting between data IntList = Empty | Cons Int IntList and data List a = Empty | Cons a (List a).
>
> Note also that if we add or remove parameters, we change a type's kind, and will need to update other typedefs which depend on it.

With https://github.com/hackworthltd/primer/issues/401 being closed, I will use this issue to also track deletion of parameters.

brprice:
@georgefst Could you update us on the current status of this?

IIUC, we support adding (and deleting) a parameter but only when the type is unused?

georgefst:
Like #401, I forgot to come back to this issue when this was partially implemented in (the somewhat mis-titled) #1063, adding support for these edits in _unused_ types.

For supporting _used_ types, we need to consider how to fix up type-level expressions where the kind of a type has changed. It _might_ be that this kind/type case is not significantly harder than the already-implemented type/term case. We just have not yet attempted to enumerate all the possible interactions in the manner of https://github.com/hackworthltd/primer/issues/267. Note that when we do so, we should consider _deletion_ of parameters (which overlaps with https://github.com/hackworthltd/primer/issues/401).

Also, note that while the opening comment talks only about adding and removing params, we also support editing the kinds of parameters (https://github.com/hackworthltd/primer/pull/1095), and need to consider how to handle these changes for in-use types.

Issue #401, opened 2022-04-19
Editable typedefs: deletion


georgefst:
From "not in spec" section of https://github.com/hackworthltd/primer/issues/267:

> The deletion of parts of a type definition is harder to handle, since we want to avoid the student losing code that they've written. In addition to the operations outlined above, the following are also possible:
>
> - Delete type
> - Delete constructor
> - Delete parameter from constructor
> - Delete parameter from type
>   - Concerns about losing code are somewhat lesser here in practice, but a student could in theory have an arbitrarily-complex type-level expression in a no-longer-required type application.
>   - Note that we can only allow this when the parameter does not appear in any constructors. Or at least, we should remove the parameter from any constructors automatically as a first step.
>
> We should consider these alongside other changes which can cause code to disappear. See [Saving the student's work when a type changes](https://github.com/hackworthltd/primer/issues/132#issuecomment-873428615).

brprice:
@georgefst Could you update us on the status of this?

IIUC, we support deletions of "unused" types/constructors/parameters-of-ctors/parameters-of-types?

georgefst:
Yes, we now support deleting parts of _unused_ typedefs. This was implemented in #1063 and got a bit lost in the Zurihac maelstrom.

The check is not as fine-grained as it could be (e.g. we should be able to delete an unused constructor even when its type is in use - https://github.com/hackworthltd/primer/pull/1063#discussion_r1235022965).

Anyway, we're now intending to just warn about more destructive deletions, instead of banning them. See https://github.com/hackworthltd/primer/issues/1133.

georgefst:
I consider #1153 to close this issue, since all that now remains is tracked in #1133.

Issue #399, opened 2022-04-19
Run typechecker whenever a typedef changes


georgefst:
There are two issues with #367 which would seemingly be solved by some variation of "run a full typechecker pass":

## Unnecessary holes

From https://github.com/hackworthltd/primer/pull/367#issue-1196027941:

> There is one part of the spec which we don't implement here: it says that we don't add unnecessary holes - see e.g. "when the program does not otherwise typecheck". But in this PR, we are cautious and always add holes where they may be needed, without examining the wider context.

There's some progress on [this branch](https://github.com/hackworthltd/primer/tree/georgefst/editable-typedefs-1-smartholes), but I clearly don't quite know what I'm doing there. The tests on that branch are certainly useful though.

### Example

Here is one example of where we currently insert unnecessary holes.

We start with:

```
data T = A Bool

d = let x = ? in A x
```

The student changes the type of the first field of `A`:

```
data T = A Int
```

And Primer inserts a hole:

```
d = let x = ? in A {? x ?}
```

Further examples are in the tests on the branch mentioned above.

## Invalid metadata

From https://github.com/hackworthltd/primer/issues/267#issuecomment-1095211670:

> > Something else which wasn't considered above: what to do with metadata? e.g. when we end up with types in metadata with the wrong name.
> > Should we just take the safe option and traverse all expressions, wiping all `TypeCache`s? We might be able to recover from simple edits like renaming, but other edits may mean that types in metadata are no longer well-formed anyway.
>
> I think the safe option is to do a full TC pass after the edit, which will re-populate all the caches. This would ensure that (a) the frontend gets up-to-date info and (b) that we will get an error message promptly if there is a bug in the action code which makes the program ill-formed.

PR #949, opened 2023-04-18
feat: Expose typedef edit actions in OpenAPI

georgefst:
This allows us to finally take advantage of #367 (with the caveats detailed in #399, #401 and #402).

This will ultimately enable us to render typedefs alongside terms in our frontend, and modify them using the same actions panel with which we manipulate terms.

See #1040 and #1041 for potential follow-ups.

### DIFF
--- a/primer/src/Foreword.hs
+++ b/primer/src/Foreword.hs
@@ -32,6 +32,8 @@ module Foreword (
   curry4,
   unsafeMaximum,
   spanMaybe,
+  adjustAtA',
+  findAndAdjustA',
 ) where
 
 -- In general, we should defer to "Protolude"'s exports and avoid name
@@ -130,6 +132,11 @@ adjustAtA n f xs = case splitAt n xs of
   (a, b : bs) -> f b <&> \b' -> Just $ a ++ [b'] ++ bs
   _ -> pure Nothing
 
+adjustAtA' :: Applicative f => Int -> (a -> f (a, z)) -> [a] -> f (Maybe ([a], z))
+adjustAtA' n f xs = case splitAt n xs of
+  (a, b : bs) -> f b <&> \(b', z) -> Just (a ++ [b'] ++ bs, z)
+  _ -> pure Nothing
+
 -- | Adjust the first element of the list which satisfies the
 -- predicate. Returns 'Nothing' if there is no such element.
 findAndAdjust :: (a -> Bool) -> (a -> a) -> [a] -> Maybe [a]
@@ -142,6 +149,11 @@ findAndAdjustA :: Applicative m => (a -> Bool) -> (a -> m a) -> [a] -> m (Maybe 
 findAndAdjustA p f = \case
   [] -> pure Nothing
   x : xs -> if p x then Just . (: xs) <$> f x else (x :) <<$>> findAndAdjustA p f xs
+
+findAndAdjustA' :: Applicative m => (a -> Bool) -> (a -> m (a, z)) -> [a] -> m (Maybe ([a], z))
+findAndAdjustA' p f = \case
+  [] -> pure Nothing
+  x : xs -> if p x then (\(x', z) -> Just . (,z) . (: xs) $ x') <$> f x else first (x :) <<$>> findAndAdjustA' p f xs
 
 -- | Change the type of an error.
 modifyError :: MonadError e' m => (e -> e') -> ExceptT e m a -> m a
--- a/primer/src/Foreword.hs
+++ b/primer/src/Foreword.hs
@@ -32,8 +32,6 @@ module Foreword (
   curry4,
   unsafeMaximum,
   spanMaybe,
-  adjustAtA',
-  findAndAdjustA',
 ) where
 
 -- In general, we should defer to "Protolude"'s exports and avoid name
@@ -132,11 +130,6 @@ adjustAtA n f xs = case splitAt n xs of
   (a, b : bs) -> f b <&> \b' -> Just $ a ++ [b'] ++ bs
   _ -> pure Nothing
 
-adjustAtA' :: Applicative f => Int -> (a -> f (a, z)) -> [a] -> f (Maybe ([a], z))
-adjustAtA' n f xs = case splitAt n xs of
-  (a, b : bs) -> f b <&> \(b', z) -> Just (a ++ [b'] ++ bs, z)
-  _ -> pure Nothing
-
 -- | Adjust the first element of the list which satisfies the
 -- predicate. Returns 'Nothing' if there is no such element.
 findAndAdjust :: (a -> Bool) -> (a -> a) -> [a] -> Maybe [a]
@@ -149,11 +142,6 @@ findAndAdjustA :: Applicative m => (a -> Bool) -> (a -> m a) -> [a] -> m (Maybe 
 findAndAdjustA p f = \case
   [] -> pure Nothing
   x : xs -> if p x then Just . (: xs) <$> f x else (x :) <<$>> findAndAdjustA p f xs
-
-findAndAdjustA' :: Applicative m => (a -> Bool) -> (a -> m (a, z)) -> [a] -> m (Maybe ([a], z))
-findAndAdjustA' p f = \case
-  [] -> pure Nothing
-  x : xs -> if p x then (\(x', z) -> Just . (,z) . (: xs) $ x') <$> f x else first (x :) <<$>> findAndAdjustA' p f xs
 
 -- | Change the type of an error.
 modifyError :: MonadError e' m => (e -> e') -> ExceptT e m a -> m a
--- a/primer/src/Primer/API.hs
+++ b/primer/src/Primer/API.hs
@@ -37,6 +37,8 @@ module Primer.API (
   viewProg,
   Prog (Prog),
   Module (Module),
+  TypeDef (TypeDef),
+  ValCon (ValCon),
   Def (Def),
   getProgram,
   getProgram',
@@ -207,10 +209,11 @@ import Primer.Log (
   PureLog,
   runPureLog,
  )
-import Primer.Module (moduleDefsQualified, moduleName, moduleTypesQualified)
+import Primer.Module (moduleDefsQualified, moduleName, moduleTypesQualifiedMeta)
 import Primer.Name qualified as Name
 import Primer.Primitives (primDefType)
-import Primer.TypeDef (ASTTypeDef (ASTTypeDef), ValCon (ValCon))
+import Primer.TypeDef (ASTTypeDef (ASTTypeDef), astTypeDefConstructors, typeDefNameHints, typeDefParameters)
+import Primer.TypeDef qualified as TypeDef
 import StmContainers.Map qualified as StmMap
 
 -- | The API environment.
@@ -627,7 +630,7 @@ data Prog = Prog
 data Module = Module
   { modname :: ModuleName
   , editable :: Bool
-  , types :: [TyConName]
+  , types :: [TypeDef]
   , -- We don't use Map Name Def as it is rather redundant since each
     -- Def carries a name field, and it is difficult to enforce that
     -- "the keys of this object match the name field of the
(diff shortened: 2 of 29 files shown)


primer/src/Primer/Action.hs
```diff
@@ -910,35 +968,72 @@ toProgActionNoInput defs def defName mNodeSel = \case
...
+      (ty, s) <- conSel
+      maybe (Left NeedTypeDefConsFieldSelection) (pure . (ty,s.con,)) s.field
+    toProgAction actions = do
+      case sel0 of
+        SelectionDef sel -> toProg' actions sel.def <$> maybeToEither NoNodeSelection sel.node
+        SelectionTypeDef _ -> do
+          (t, c, f) <- conFieldSel
+          pure [ConFieldAction t c f.index $ SetCursor f.meta : actions]
+    termDef = first (const NeedTermDef) def0
+    typeDef = either Right (Left . const NeedTypeDef) def0
```
georgefst, line 1026: A lot of these helpers also occur in `toProgActionInput`. I started trying to clean this up in #1041, but got in to the weeds a little. I'd prefer that we avoid this repetition for this PR, but revisit soon.

primer/src/Primer/Action.hs
```diff
@@ -230,6 +244,49 @@ applyActionsToTypeSig smartHoles imports (mod, mods) (defName, def) actions =
...
+      wrappedType <- ann emptyHole (pure ty)
+      let unwrapError = throwError $ InternalFailure "applyActionsToField: failed to unwrap type"
+          wrapError = throwError $ InternalFailure "applyActionsToField: failed to wrap type"
+          focusedType = focusType $ focus wrappedType
+      case focusedType of
+        Nothing -> wrapError
+        Just wrappedTy ->
+          f wrappedTy >>= \case
+            InType zt -> pure zt
+            z -> maybe unwrapError pure (focusType (unfocusLoc z))
```
georgefst, line 288: This whole function is a bit of a copy-paste from `applyActionsToType`, and I'm not certain that all details are relevant, or whether there's anything missing. Also, I don't know if there's anything from the recently-fixed `SetConFieldType` (#932) that we're missing. However, #1040 doesn't seem to be revealing any issues from `forTypeDefConsFieldNode`, so maybe we're in the clear.

primer/src/Primer/Action.hs
```diff
@@ -230,6 +244,49 @@ applyActionsToTypeSig smartHoles imports (mod, mods) (defName, def) actions =
...
+    & runExceptT
+  where
+    go :: ActionM m => m ([Module], TypeZ)
+    go = do
+      (valCons, zt) <-
+        (maybe (throwError $ InternalFailure "applyActionsToField: con name not found") pure =<<) $
+          flip (findAndAdjustA' ((== conName') . valConName)) (astTypeDefConstructors tyDef) \(ValCon _ ts) -> do
+            (t, zt) <-
+              maybe (throwError $ InternalFailure "applyActionsToField: con field index out of bounds") pure
+                =<< flip (adjustAtA' index) ts \fieldType -> do
```
georgefst, line 268: I don't really like `adjustAtA'` and `findAndAdjustA'`. I feel like there _should_ be some way to use an appropriate `Applicative`, like `Compose (,)`, rather than having to create a whole new helper to thread the extra data along. But I couldn't work out how to make this work. Any ideas?

primer/src/Foreword.hs
```diff
@@ -143,6 +150,11 @@ findAndAdjustA p f = \case
   [] -> pure Nothing
   x : xs -> if p x then Just . (: xs) <$> f x else (x :) <<$>> findAndAdjustA p f xs
 
+findAndAdjustA' :: Applicative m => (a -> Bool) -> (a -> m (a, z)) -> [a] -> m (Maybe ([a], z))
+findAndAdjustA' p f = \case
+  [] -> pure Nothing
+  x : xs -> if p x then (\(x', z) -> Just . (,z) . (: xs) $ x') <$> f x else first (x :) <<$>> findAndAdjustA' p f xs
```
georgefst, line 156: _If_ these functions are staying (see above), I should at least add haddocks, and maybe pick a better name.

primer/src/Primer/Action.hs
brprice: > I'd prefer that we avoid this repetition for this PR, but revisit soon.

As in: let's just commit this as-is, and improve it in #1041?

georgefst: Ah, yes, I seemed to be missing a few words.

primer/src/Primer/API.hs
```diff
@@ -633,6 +636,25 @@ data Module = Module
   deriving (ToJSON, FromJSON) via PrimerJSON Module
   deriving anyclass (NFData)
 
+data TypeDef = TypeDef
+  { name :: TyConName
+  , params :: [TyVarName]
+  , nameHints :: [Name.Name]
+  , constructors :: Maybe [ValCon]
+  -- ^ a `Nothing` here indicates a primitive type (whereas `Just []` is `Void`)
+  }
```
brprice, line 655: Why do these need name hints? Just so a frontend can display them if wanted?

primer/src/Primer/App/Base.hs
```diff
@@ -54,21 +63,63 @@ data NodeType = BodyNode | SigNode
   deriving (FromJSON, ToJSON) via PrimerJSON NodeType
   deriving anyclass (NFData)
 
--- | Describes what interface element the user has selected.
--- A definition in the left hand nav bar, and possibly a node in that definition.
+-- | Describes which element of a (type or term) definition the student has selected.
+-- We have the following invariant: when this contains a `NodeSelection` with @nodeType = SigNode@,
+-- or any `TypeDefConsFieldSelection`, then they will always have @meta = Right _@.
```
brprice, line 72: Markup seems odd here. `s/`/@/g`?

primer/src/Primer/JSON.hs
```diff
@@ -43,4 +42,10 @@ import Deriving.Aeson.Stock (Vanilla)
 --
 -- * @SumTwoElemArray@ is unsupported by openapi3 as it is unrepresentable in
 --   a schema.
-type PrimerJSON a = Vanilla a
+type PrimerJSON a =
+  CustomJSON
+    '[ -- This protects some clients (e.g. TypeScript) from spurious equality failures
+       -- due to null-vs-omitted inconsistency.
+       OmitNothingFields
```
brprice, line 49: In #966 etc, there was issues with `OmitNothingFields`. I assume that they are solved now that ghc has been upgraded? It may be good to keep this in mind in case there is a regression in the future

primer/src/Primer/Action.hs
brprice: r.e. #932, I think this is fine: you are calling `checkEverything` instead of manually fixing all usages of this constructor.

There is some DRYing that could happen to `withWrappedType` in particular, but this could be done later.

primer/src/Primer/App/Base.hs
georgefst: I think the backtick is fine, isn't it? We use it the same way elsewhere. And the `@`'s are there (actually just moved from another file) because they enclose a compound expression, rather than a single identifier.

primer/src/Primer/App.hs
```diff
@@ -813,9 +824,19 @@ applyProgAction prog mdefName = \case
...
+      ms' <- traverseOf (traversed % #moduleDefs) updateDefs (m' : ms)
+      pure
+        ( ms'
+        , Just
+            . SelectionTypeDef
+            . TypeDefSelection type_
+            . Just
+            . TypeDefConsNodeSelection
+            . TypeDefConsSelection con
+            $ Nothing
```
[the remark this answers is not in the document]
georgefst: Yes

primer/src/Primer/App.hs
```diff
@@ -522,7 +540,8 @@ handleEditRequest actions = do
     go :: (Prog, Maybe GVarName) -> ProgAction -> m (Prog, Maybe GVarName)
     go (prog, mdef) a =
       applyProgAction prog mdef a <&> \prog' ->
-        (prog', (.def) <$> progSelection prog')
+        -- TODO this ignores any selection from a type def selection
```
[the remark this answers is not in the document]
georgefst: Ah, yes. This is because of https://github.com/hackworthltd/primer/pull/964. I'll work out what to do with that PR, probably either merging it first or closing it and cherry-picking here.

primer/src/Primer/Action.hs
brprice: As things stand, there are two difficulties.

Firstly, from the type of `adjustAtA`, we have no reason to believe that the function will be called any particular number of times, thus whatever `Applicative` we use needs to handle the possibility of multiple `z`s. There are three ways to improve this: either weaken the `Applicative` constraint to something like `Pointed` (but this steps out of the standard hierarchy, so may be awkward to use); define something like `adjustAtA0 :: Functor f => Int -> (a -> f a) -> [a] -> Maybe (f [a])` (essentially the same idea, but makes it explicit in the return type; this new function can be used to define `adjustAtA`); or keep the current type and use `First`.

The second difficulty is the chosen order of the tuple. We have no way to say `(,z)` at type level, only `(z,)`, written as `(,) z`. If you change the order of the tuple it is then possible to define `adjustAtA'` in terms of `adjustAtA`.

This boils down to one of the following:
```
adjustAtA0 :: Functor f => Int -> (a -> f a) -> [a] -> Maybe (f [a])
adjustAtA0 n f xs = case splitAt n xs of
  (a, b : bs) -> Just $ f b <&> \b' -> a ++ [b'] ++ bs
  _ -> Nothing

-- | Like 'adjustAt', but in an 'Applicative'.
adjustAtA :: Applicative f => Int -> (a -> f a) -> [a] -> f (Maybe [a])
adjustAtA n f xs = sequenceA $ adjustAtA0 n f xs

adjustAtA' :: forall f a z . Applicative f => Int -> (a -> f (z, a)) -> [a] -> f (Maybe (z, [a]))
adjustAtA' n f = traverse getCompose . adjustAtA0 n f'
  where f' :: a -> Compose f ((,) z) a
        f' = Compose . f
```

```
adjustAtA' :: forall f a z . Applicative f => Int -> (a -> f (z, a)) -> [a] -> f (Maybe (z, [a]))
adjustAtA' n f = fmap (_ . first getFirst) . getCompose . adjustAtA n f'
  where f' :: a -> Compose f ((,) (First z)) a
        f' = Compose . fmap (first (First . Just)) . f
        g :: Applicative g => (g a, g b) -> g (a, b)
        g (x,y) = liftA2 (,) x y
```

brprice (approved):
Other than the issue with `genNames`, this looks good to me

primer/src/Primer/API.hs
dhess: Yeah, let's leave these in place. It would be great if students could add their own hints.

primer/src/Primer/Action.hs
georgefst: > r.e. #932, I think this is fine: you are calling `checkEverything` instead of manually fixing all usages of this constructor.

Ah, of course. 

And indeed, given how general one of these actions can be, there didn't seem much alternative. But I suppose, now that you allude to it, we could in theory look for uses of the given constructor, and try to fix things more locally. I'm certainly not suggesting that we change that in this PR, however.

primer/src/Primer/Action/ProgAction.hs
```diff
@@ -47,6 +47,8 @@ data ProgAction
     BodyAction [Action]
   | -- | Execute a sequence of actions on the type annotation of the definition
     SigAction [Action]
+  | -- | Execute a sequence of actions on the type of a field of a constructor in a typedef
+    ConFieldAction TyConName ValConName Int [Action]
```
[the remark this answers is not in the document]
georgefst: Removed.

primer/src/Primer/App.hs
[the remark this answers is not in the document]
georgefst: Rebased on said PR.

primer/src/Primer/Action/Available.hs
```diff
@@ -461,30 +472,47 @@ options typeDefs defs cxt level def0 sel0 = \case
...
+    genNames typeOrKind =
+      map localOpt . flip runReader cxt <$> case sel0 of
+        SelectionDef sel -> do
+          z <- focusNode =<< sel.node
+          pure $ case z of
+            Left zE -> generateNameExpr typeOrKind zE
+            Right zT -> generateNameTy typeOrKind zT
+        SelectionTypeDef sel -> do
+          (_, zT) <- conField sel
+          pure $ generateNameTy typeOrKind zT
```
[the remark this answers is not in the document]
georgefst: Fixed in final commit.

PR #966, opened 2023-04-27
Selda bug on GHC 9.2.7 (with other changes from GHC 9.4.5 branch)

dhess:
Note to selves: this is a repro for the Selda bug that @georgefst found. To summarize:

* #965, with GHC 9.4.5, works.
* This PR, which is nearly identical to #965 except that it uses GHC 9.2.7, fails (as in, the selda test runs forever at 100% CPU).


dhess:
I'm closing this, as it's not relevant to our repo anymore.

brprice:
> I'm fairly confident that this is an aeson/ghc bug: the part that chews cpu is in serialising a value to json (and it runs swiftly with either (a) not having `OmitNothingFields` or (b) "good" versions of ghc iirc these include 8.10.7, 9.0.* and 9.4.5, but not 9.2.* or 9.4.{1..4}). When I get some time I'll open an upstream issue (I have a very non-minimal repro showing the issue just depending on aeson) and link it here.

Upstream issue is at https://github.com/haskell/aeson/issues/1030

PR #973, opened 2023-05-09
DNM: messy test to try to find undo bug; only found some other bugs

brprice:
I finally hacked together the test I have mentioned to try to flush out any undo bugs: "run a bunch of actions, then ensure can undo at the end". This did not uncover any bugs. Thus I conclude that either
- the test is inadequate: perhaps some interaction (other than taking an action listed in the test) can tick the fresh counter but not alter the log, thus when a future action is taken it has a higher ID than if the "silent" action was not taken, thus breaking replays (i.e. later undos)
- or @georgefst's hunch that the only bug is related to migrations and will only show up if we change the implementation in such a way as to change generated IDs and then try to "undo" in a saved session that was created before that change.

I am mostly opening this PR so the code is not lost. In the case that this bug does crop up again, this may be a good starting point to find it. I'm happy to do the (fairly minor) clean up to get this into a mergeable state if you think it would be a good addition to the testsuite, but I doubt I'll push for its inclusion myself.

### DIFF
--- a/primer/gen/Primer/Gen/App.hs
+++ b/primer/gen/Primer/Gen/App.hs
@@ -8,6 +8,7 @@
 module Primer.Gen.App (
   genProg,
   genApp,
+  extendCxtByModules,
 ) where
 
 import Control.Monad.Fresh (MonadFresh (fresh))
@@ -88,10 +89,6 @@ genProg sh initialImports = local (extendCxtByModules initialImports) $ do
           pure $ x : rest
     extendCxtByModule :: Module -> Cxt -> Cxt
     extendCxtByModule = extendCxtByModules . pure
-    extendCxtByModules :: [Module] -> Cxt -> Cxt
-    extendCxtByModules ms =
-      extendTypeDefCxt (foldMap' moduleTypesQualified ms)
-        . extendGlobalCxt (M.toList . fmap (forgetTypeMetadata . defType) $ foldMap' moduleDefsQualified ms)
     genModule :: Name -> Int -> GenT WT Module
     genModule prefix index = do
       let mn = ModuleName $ prefix :| [unsafeMkName $ show index]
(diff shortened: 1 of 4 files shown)


brprice:
I have diagnosed one bug here (I am unsure if there are multiple ones lurking).
However, this bug, similarly to the one in the gist linked earlier actually fails *before* we test undo, i.e. it is a bug in offered-actions-being-accepted! Thus, this is not actually an undo bug.

 Sometimes we find a situation where 71c86a469527fced60e363405ddbbd1ac0cacdec "feat: actions handle smart elision of hole/ann better" is incomplete, such as the following minimised example
```
unit_tmp_5 :: Assertion
unit_tmp_5 =
  let ((builtins, d), i) = create $ do
        m <- builtinModule
        astDefExpr <- hole $ emptyHole `ann` tcon tBool
        astDefType <- tcon tNat
        pure (m,ASTDef
               { astDefExpr
               , astDefType
               })
  in case evalTestM i $ applyActionsToBody SmartHoles [builtins] d [Action.Move Child1, Action.EnterType, Action.Delete] of
    Left err -> assertFailure $ show err
    Right _ -> pure ()
```
this starts with `foo : Nat ; foo = {?  ? : Bool ?}`, focuses on the `Bool` and issues `Delete` which gives `foo = {? ? : ? ?}` and then smartholes gives `foo = ?` Note that the ID of the remaining hole is the same as the original empty hole. Our fix in 71c86a469527fced60e363405ddbbd1ac0cacdec only deals with what happens if we are focused on an annotation node which is smartholes-elided, and not if we are focussed on the type child of an annotation node!

brprice:
Hmm, I am rather confused about this test. It uncovered some bugs that were fixed in #1032 , but none of these were related to undo! I do not understand why it seems a more effective test than the current "offered actions are accepted" test, though I have not looked into it closely. It may be worth (tidying up and) merging just for the more effective test, even if it does not tickle an undo bug.

brprice:
Given that we now have understood at least one undo bug #1056 which can only be triggered by doing multiple interleaved edit, eval, undo and redo actions, I think it would be worth getting this test (upgraded? and) merged, as our current test only does one action, starting from a complex generated App

dhess:
Can this be closed now?

dhess:
Bumping, can this PR be closed?

brprice:
Closing in favour of #1167

PR #986, opened 2023-05-13
refactor: DRY repeated `meta`, `TypeZip'` definitions

brprice:


brprice:
> Out of curiosity, any idea how this happened?

For `type TypeZip' = ...`, it happened in 08b6e3aeb1a33bcf27e70febd9440dd353d259d4 "refactor: split Zipper into bits that only depend on Type": I guess I simply missed moving that definition, instead cloning it. Since it is only a synonym, having the same definition in two files did not cause GHC to complain.

For `meta`, it happened in 4e8f1ff55bc17791992fb71784e466e36a87961a "refactor: split DSL to unlock trimming of module dependencies". I guess it was a similar copy/paste error. However, this happened to increase parallelism in the module graph; I suspect that this will not impact build times, since it only actually affects whether `Primer.Core.DSL.{Meta,Type}` can be built in parallel with each other or not, and they are both tiny modules.

PR #1009, opened 2023-05-20
chore: build HLS from haskell.nix again

dhess:
HLS 2.0.0.0 is out, let's try again.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -365,23 +365,6 @@
         "type": "github"
       }
     },
-    "haskell-language-server": {
-      "flake": false,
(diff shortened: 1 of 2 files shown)


georgefst:
From a glance at the build log, maybe dry disabling the stylish-haskell plugin? Since we don't use that formatter anyway.

dhess:
Yeah, though that means going back to the custom build again rather than the `haskell.nix` built-in, which was counter to the point of this PR 😭

dhess:
Cool, that fix seems to have done the trick.

PR #1014, opened 2023-05-20
chore: add a flake config for our public Cachix cache

dhess:


dhess:
After playing with this a bit, I think it's a bad idea. It appears to have some issues with `direnv`, our Nix shell CI caching script gets stuck when this config is included, and the `--no-accept-flake-config` option, which I expected would un-stick the script, doesn't appear to do anything.

Instead, I'll add docs on how to enable our public Cachix cache the usual way(s).

PR #1021, opened 2023-05-21
ci: push benchmarks to `cloudflare-pages` branch

dhess:
We'll use this branch to publish benchmarks to Cloudflare Pages,
rather than GitHub Pages.


dhess:
Turns out this isn't necessary. We can keep the current configuration here in the repo; all the necessary config is in Cloudflare.

PR #1026, opened 2023-05-22
doc: clean up `Foreword` Haddocks

dhess:

### DIFF
--- a/primer/src/Foreword.hs
+++ b/primer/src/Foreword.hs
@@ -1,11 +1,23 @@
 {-# LANGUAGE ExplicitNamespaces #-}
 
+-- |
+-- Module: Foreword
+-- Description: Primer's custom prelude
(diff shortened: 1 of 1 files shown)


dhess:
Cleaning up Haddocks is a goal for pre-1.0, and the sooner the better, but it's probably going to take awhile, and will be tedious. I'm not sure yet whether I'll just do bits here and there as a bunch of PRs at various times, or whether I'll bunch them up in large batches. It'll probably be a mix of both.

Before I embark on widespread formatting changes, it would be helpful if we had a way to see the generated Haddocks for any given PR. This will take a bit of work, but should be achievable without jumping through too many hoops, especially if we can use something like GitHub Pages or Cloudflare Pages integration with GitHub workflows. Unfortunately, based on our previous experience trying to coordinate Nix+Buildkite builds with triggered GitHub workflows, this is where some friction will occur.

PR #1029, opened 2023-05-22
fix: letrec metadata

brprice:
Consider (some sort of) a `let` in a checkable position, whose body is
synthesisable, e.g. `(λx. let y = ... in xor x y) : Bool -> Bool`. Prior to
this commit, our treatment of metadata between `let` and `letrec` was
inconsistent:
- they both treat the body `xor x y` the same: checking it against
  `Bool`, finding it is synthesises `Bool` and then giving it a `TCEmb`
  metadata.
- a `let` node would then get the same metadata
- a `letrec` node would only get `TCChkedAt` metadata

This is visible to a consumer of our API: what actions we offer differ,
see the added test.

### DIFF
--- a/primer/src/Primer/Action.hs
+++ b/primer/src/Primer/Action.hs
@@ -10,6 +10,8 @@ module Primer.Action (
   applyActionsToTypeSig,
   applyActionsToExpr,
   moveExpr,
+  enterType,
+  moveType,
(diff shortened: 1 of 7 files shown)


brprice:
This is a clone of #979, which was accidentally closed.

PR #1049, opened 2023-05-30
feat: wildcards

brprice:


brprice:
Status: the final three commits are comments that I wanted to bring up in review. The rest is pretty much ready to go, simply needing better commit messages (and to be `--sign-off`-ed)

primer/src/Primer/Core/DSL.hs
```diff
@@ -159,6 +160,10 @@ letType v t e = LetType <$> meta <*> pure v <*> t <*> e
 case_ :: MonadFresh ID m => m Expr -> [m CaseBranch] -> m Expr
 case_ e brs = Case <$> meta <*> e <*> sequence brs <*> pure CaseExhaustive
 
+-- | A non-exhaustive case
+caseFB_ :: MonadFresh ID m => m Expr -> [m CaseBranch] -> m Expr -> m Expr
+caseFB_ e brs fb = Case <$> meta <*> e <*> sequence brs <*> (CaseFallback <$> fb)
```
brprice: We leave `case_` alone to avoid churn. (I should mention this in the commit message)

primer/src/Primer/API.hs
```diff
@@ -829,6 +829,29 @@ viewTreeExpr e0 = case e0 of
...
+            patternRootId = boxId <> "B"
+           in
+            Just $
+              Tree
+                { nodeId = boxId
+                , body =
+                    BoxBody . RecordPair Flavor.Pattern $
+                      ( Tree
+                          { nodeId = patternRootId
+                          , body = TextBody $ RecordPair Flavor.PatternWildcard (Name Nothing "_")
```
brprice: It is somewhat odd to render the wildcard match as a text "_" node...
Any better ideas?

primer/src/Primer/Action.hs
```diff
@@ -733,6 +741,58 @@ constructCase ze = do
         TC.SmartHoles -> flip replace ze <$> case_ (pure $ target ze) []
     _ -> throwError $ CustomFailure ConstructCase ("can only construct case on expression with type a exactly saturated ADT, not " <> show ty)
 
+addCaseBranch :: ActionM m => QualifiedText -> ExprZ -> m ExprZ
+addCaseBranch rawCon ze = case target ze of
```
brprice: In general, all our "add/delete case branches" actions (both high and low level) are (will) be offered/accepted on the `case` node itself. This feels a bit odd but is I think the best we can do for now, since we don't have the ability to select one branch -- we can only select things in branches or binders in patterns. A different choice would be to give "add" at the root of the wildcard RHS and "delete" at the root of each non-wildcard branch's rhs

primer/src/Primer/Action/Available.hs
```diff
@@ -246,13 +251,14 @@ forExpr tydefs l expr =
         <> annotate
         <> [Input RenameLet]
     Case _ scrut brs fb ->
```
brprice, line 253: TODO: we don't actually offer `MakeCase` on a primitive

primer/src/Primer/API.hs
dhess: We also considered not rending the pattern box at all in this case and going straight into the output. However, we realized that while this might be the nicest solution for the wildcard pattern, it wouldn't mesh well with deep patterns (once we can support them) that have nested wildcard patterns.

brprice: We now render it as its own flavor with no body

primer/src/Primer/Action/Available.hs
brprice: fixed

PR #1051, opened 2023-05-30
Display types as leftmost child to match convention for top-level definitions

georgefst:
@dhess

### DIFF
--- a/primer/src/Primer/API.hs
+++ b/primer/src/Primer/API.hs
@@ -708,7 +708,7 @@ viewTreeExpr e0 = case e0 of
     Tree
       { nodeId
       , body = NoBody Flavor.Ann
-      , childTrees = [viewTreeExpr e, viewTreeType t]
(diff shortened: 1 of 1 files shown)


georgefst:
We discussed this in yesterday's meeting.

We decided that it would be rather weird and unconventional to display the type first in text mode, and we'd like to retain a symmetry between tree and text mode. For one thing, we might eventually want an animation transforming between the two, and we'd like to keep this as simple as possible.

(Also, subtle technicality: we need to be careful how we justify this PR, since it's actually motivated by the behaviour of our frontend. The backend (this repository) outputs the type and term of top-level definitions separately, and its our frontend that chooses to display them as part of a single tree with the type on the left. This may actually change, since we're considering moving to a more windowed UI, rather than displaying all definitions on a single canvas, in which case the motivation for displaying all parts of a definition as a single tree would be much weaker.)

PR #1053, opened 2023-06-01
test: adjust distributions to make duplicates more likely

brprice:
We already did this for `genName`.

Signed-off-by: Ben Price <ben@hackworthltd.com>


brprice:
Whilst working on #1049, I noticed that our generators would not (i.e. it was technically possible, but vanishingly small probability) generate the same primitive twice (which would make the "available actions accepted" test fail, due to adding an existing branch). This change in the generators should make it somewhat likely.

dhess (approved):

PR #1055, opened 2023-06-05
feat: optionally do not evaluate under binders (except lets)

brprice:
This change means our EvalFull is "closed evaluation", rather than "open evaluation". The main reason to do this is so that evaluations of programs-in-construction (especially recursive ones) do not pointlessly blow up in size: if `even` and `odd` are defined recursively, then evaluating `even`  would evaluate under the lambda and inside case branches, expanding `odd` and recursing; when it eventually times out one would have a big tree with many unrolled copies of `even` and `odd`, which is not very illuminating. 

Note that we do not evaluate in the RHS of pattern match branches which bind variables, for the same reason as we do not evaluate under lambdas; for consistency we also do not evaluate in the RHS of branches which do not bind variables.

### DIFF
--- a/primer-api/src/Primer/API.hs
+++ b/primer-api/src/Primer/API.hs
@@ -226,7 +226,7 @@ import Primer.Def (
   defAST,
  )
 import Primer.Def qualified as Def
-import Primer.Eval (NormalOrderOptions (UnderBinders))
+import Primer.Eval (NormalOrderOptions (StopAtBinders))
 import Primer.Eval.Redex (Dir (Chk), EvalLog)
 import Primer.EvalFull (TerminationBound)
 import Primer.JSON (
@@ -1142,7 +1142,7 @@ evalFull' = curry4 $ logAPI (noError EvalFull') $ \(sid, lim, closed, d) ->
             { evalFullReqExpr = e
             , evalFullCxtDir = Chk
             , evalFullMaxSteps = fromMaybe 10 lim
-            , evalFullOptions = fromMaybe UnderBinders closed
+            , evalFullOptions = fromMaybe StopAtBinders closed
             }
       pure $ case x of
         App.EvalFullRespTimedOut e' -> EvalFullRespTimedOut $ viewTreeExpr e'
--- a/primer-api/src/Primer/API.hs
+++ b/primer-api/src/Primer/API.hs
@@ -226,6 +226,7 @@ import Primer.Def (
   defAST,
  )
 import Primer.Def qualified as Def
+import Primer.Eval (NormalOrderOptions (UnderBinders))
 import Primer.Eval.Redex (Dir (Chk), EvalLog)
 import Primer.EvalFull (TerminationBound)
 import Primer.JSON (
@@ -412,7 +413,7 @@ data APILog
   | GenerateNames (ReqResp (SessionId, ((GVarName, ID), Either (Maybe (Type' ())) (Maybe (Kind' ())))) (Either ProgError [Name.Name]))
   | EvalStep (ReqResp (SessionId, EvalReq) (Either ProgError EvalResp))
   | EvalFull (ReqResp (SessionId, EvalFullReq) (Either ProgError App.EvalFullResp))
-  | EvalFull' (ReqResp (SessionId, Maybe TerminationBound, GVarName) EvalFullResp)
+  | EvalFull' (ReqResp (SessionId, Maybe TerminationBound, Maybe NormalOrderOptions, GVarName) EvalFullResp)
   | FlushSessions (ReqResp () ())
   | CreateDef (ReqResp (SessionId, ModuleName, Maybe Text) Prog)
   | CreateTypeDef (ReqResp (SessionId, TyConName, [ValConName]) Prog)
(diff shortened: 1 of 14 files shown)


brprice:
Status: seems to work, and the approach seems correct, but not tested. The testsuite has not been updated. I would be somewhat unhappy merging this implementation, as it should at least be made configurable (perhaps separate toggles for "eval under lambda", "eval in case branches that bind" and "eval in case branches which have no binders")

primer-api/src/Primer/API.hs
```diff
@@ -1128,6 +1129,7 @@ evalFull' = curry3 $ logAPI (noError EvalFull') $ \(sid, lim, d) ->
             { evalFullReqExpr = e
             , evalFullCxtDir = Chk
             , evalFullMaxSteps = fromMaybe 10 lim
+            , evalFullOptions = UnderBinders -- TODO/REVIEW: should we expose this to the actual api? If so, the BREAKING CHANGE commit message needs updating
```
brprice, line 1132: Thoughts? Should the option be exposed to the OpenAPI for EvalFull?

primer/src/Primer/Eval/NormalOrder.hs
```diff
@@ -129,12 +131,17 @@ data RedexWithContext
...
+viewLet dez@(_, ez) = case (target ez, exprChildren' dez) of
   (Let _ x e _b, [_, bz]) -> Just (LetBind x e, bz)
   (Letrec _ x e ty _b, [_, bz]) -> Just (LetrecBind x e ty, bz)
   (LetType _ a ty _b, [bz]) -> bz `seq` Just (LetTyBind $ LetTypeBind a ty, bz)
   _ -> Nothing
 
+data NormalOrderOptions -- TODO/REVIEW: this is a bit of a misnomer, as it affects `Eval.redexes`, which is only used for single-step. NB: even if we set StopAtBinders, we can still call `Eval.step` with the id of a redex under a binder and it will reduce, even though `redexes` will not report that id
+  = UnderBinders
+  deriving stock (Eq, Show, Read, Generic)
+  deriving (FromJSON, ToJSON) via PrimerJSON NormalOrderOptions
```
brprice: Thoughts? I would like either a better name, or to decide that `redexes` will always allow eval under binders. Probably worth thinking about what API consumers would prefer here.

primer-api/src/Primer/API.hs
dhess: Yes. Unless I'm missing something, that's the point of this PR, no?

primer/src/Primer/Eval/NormalOrder.hs
dhess: Unless I misunderstand this option, I don't think we'll want to prevent eval under binders in step-eval.

primer-api/src/Primer/API.hs
brprice: > Yes. Unless I'm missing something, that's the point of this PR, no?

Well, we could always hardcode one particular choice (probably `StopAtBinders`, else this PR is somewhat non-useful). I suppose the question should better be phrased as:

"Should we hardcode `StopAtBinders`, or expose the option to the OpenAPI?"

primer/src/Primer/Eval/NormalOrder.hs
brprice: The option can be set per API call. So the question is "should step-eval be hardcoded to one choice (either allow eval under binders or not), or should the consumer decide?". (Actually, the way it is implemented is `redexes` has the option, and simply reports which nodes are redexes; a frontend is expected to ask for one of these to be reduced in a separate API call, and this call is ignorant of the option -- technically the frontend can ask for redexes-disallowing-eval-under-binders, and then request to reduce some other node which happens to be under a binder and it will just work. However, this is a somewhat pointless technical aside!)

primer-api/src/Primer/API.hs
dhess: Ahh, I see. There's no harm in exposing it, and it might be helpful at least to work around any bugs we might encounter related to this feature, but the default value should probably be `True`.

primer/src/Primer/Eval/NormalOrder.hs
dhess: I can't come up with any reasons why we'd want to disable eval under binders in step eval. Am I missing anything?

primer-api/src/Primer/API.hs
brprice: Agree. I've pushed a new commit exposing it.

primer/src/Primer/Eval/NormalOrder.hs
dhess: This option is only currently intended to prevent runaway eval of top-level functions in the eval window. I feel like we’re talking past each other because I don’t think understand why you think it would have any value in step mode. Let’s discuss on our next face-to-face meaning so we don’t go in circles here.

brprice: We decided that step eval should always be able to go under binders. We feel that it is less confusing to allow students to evaluate every redex at the expense of full-eval consistency. (Recall that the impetus for this PR is to avoid run-away inlining of recursive top-level-definitions, rather than a philosophical stance on what a "normal form" should be)

TODO
- [ ] Implement this change

dhess:
Once #736 is merged, I would like to get this merged, as well. This will become even more important once we implement #1119.

brprice:
This has been rebased on top of the push-down-lets work, and is ready for a review.

Reviewer note: please give extra scrutiny to comments and code that may have not been updated for push-down-lets (this branch originated before #736)

dhess:
@brprice Can you mark this as ready for review so that it shows up in the UI?

dhess:
> For the record, I'm still not fully convinced that this is the right long-term solution to the UX issue of showing unproductive eval steps for unapplied functions etc. But since we're making this optional that doesn't matter.

Any better ideas?

georgefst:
I'm sure we discussed this at length at some point but I can't find any notes. But I remember a suggestion that we stop with an "are you sure you want to continue?"-style message when the size of the eval tree reaches a certain multiple of the original. Anyway, I don't feel too strongly about it and think further explorations should at least wait for https://github.com/hackworthltd/primer/issues/1119.

PR #1058, opened 2023-06-07
feat: expose type of current selection via OpenAPI

dhess:
This is a rebase/rewrite of #1050

### DIFF
--- a/primer-service/exe-server/Main.hs
+++ b/primer-service/exe-server/Main.hs
@@ -386,6 +386,7 @@ instance ConvertLogMessage PrimerErr LogMsg where
   convert (ApplyActionError as e) = LogMsg $ show (as, e)
   convert (UndoError e) = LogMsg $ show e
   convert (RedoError e) = LogMsg $ show e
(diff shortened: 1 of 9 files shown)


georgefst:
I've taken over rebasing this due to my familiarity with the `Selection` type etc. It should now be merge-able.

I've ended up changing the new API quite substantially, in line with https://github.com/hackworthltd/primer/pull/1050#discussion_r1221546055. This was the easiest solution, since we end up needing extra info that isn't directly available in the selection metadata, even for term definition name nodes (which, as it happens, we previously returned `Nothing` for, which would have been an issue for the frontend), and also we don't actually currently have a way to set a `SelectionTypeDef` via `ProgAction` (there's a bit of a weird inconsistency between type and term def actions, in that the former always take a selection, and the latter start from the current one - we should revisit this at some point).

I also made all of the previous `Nothing` cases throw errors directly, so the workaround implemented in https://github.com/hackworthltd/primer/pull/1050/commits/288510691997154cb5e48bf15d8462a74c05bd33 is no longer necessary.

PR #1059, opened 2023-06-07
fix: change `getTypeOrKind` OpenAPI endpoint from `GET` to `POST`

dhess:
Though this method is technically a `GET`, we've found that OpenAPI
clients often don't support complex bodies in `GET` methods, and this
one requires that we pass a `Selection`. So, just as with the
`getAvailableActions` OpenAPI endpoint, we make `getTypeOrKind` a faux
`POST`.


georgefst:
Ah, my bad, I'd forgotten about this issue with `GET`s.

PR #1066, opened 2023-06-10
feat: annotate offered valcons that match the selection's type

dhess:
This is an extension of https://github.com/hackworthltd/primer/pull/1045, which itself was a proof of concept/best effort to annotate variables that match the type of a hole up to alpha. Here we extend this feature to non-primitive value constructors.

This implementation uses `valConType` to extract the value constructor's type. For the record, the documentation for `conInfo`, which also uses `valConType`, reads as follows:

> the type an eta-expanded version of this constructor would check against e.g. `Cons`'s "type" is `∀a. a -> List a -> List a`.

### DIFF
--- a/primer/src/Primer/Action/Available.hs
+++ b/primer/src/Primer/Action/Available.hs
@@ -29,7 +29,10 @@ import Data.List ((\\))
 import Data.Map qualified as Map
 import Data.Set qualified as Set
 import Data.Text qualified as T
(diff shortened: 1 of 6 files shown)


dhess:
Future work should offer any value constructor whose type matches the hole, regardless of its arity.

georgefst (approved):

PR #1067, opened 2023-06-10
feat: better valcon type matching

dhess:
Here we ignore the arity of the constructor and consider only the type to which the constructor belongs. This matches, e.g., both `Zero` and `Succ i` to a node whose type is `Nat`; whereas the previous implementation, which eta-expands the value constructor's type, would only match `Zero`.

**However**, the current implementation of this is not quite what we want, because the way I've done this, we've lost the ability to match against any polymorphic type. For example, if we have a hole of type `List Int`, we no longer get a match on either `Cons` *or* `Nil`.

I've taken this as far as I can for now due to current time constraints, so I'll mark this as a draft for now.

### DIFF
--- a/primer/src/Primer/Action/Available.hs
+++ b/primer/src/Primer/Action/Available.hs
@@ -83,6 +83,9 @@ import Primer.Core (
   _type,
   _typeMetaLens,
  )
+import Primer.Core.Transform (
+  mkTAppCon,
+ )
 import Primer.Core.Utils (forgetTypeMetadata, freeVars)
 import Primer.Def (
   ASTDef (..),
@@ -104,7 +107,6 @@ import Primer.TypeDef (
   TypeDefMap,
   ValCon (valConArgs),
   valConName,
-  valConType,
  )
 import Primer.Typecheck (
   Cxt,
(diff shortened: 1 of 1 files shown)


primer/src/Primer/Action/Available.hs
```diff
@@ -586,7 +588,14 @@ options typeDefs defs cxt level def0 sel0 = \case
...
                       vcs <&> \vc@(vc', tc, td) ->
-                        (globalOpt' (t `eqType` valConType tc td vc') (valConName vc'), vc)
+                        -- Here we ignore the arity of the constructor
+                        -- and consider only the type to which the
+                        -- constructor belongs. This matches, e.g.,
+                        -- both @Zero@ and @Succ i@ to a node whose
+                        -- type is @Nat@; whereas using 'valConType',
+                        -- which eta-expands the value constructor's
+                        -- type, would only match @Zero@.
+                        (globalOpt' (t `eqType` mkTAppCon tc (TVar () . fst <$> astTypeDefParameters td)) (valConName vc'), vc)
```
brprice, line 598: I think we can do better here. (See below)

Background: we only have regular types (i.e. not gadts), so for a type `D a b` with a constructor `C A[a,b]`, we have that `D S T  ∋ C A[S,T]` (with no restrictions on what `S` and `T` (the instantiations of the type's params) are). Thus we only need to check whether the constructor `vc` is a constructor of `D` (and that `D S T` is fully-applied, though this should be guarenteed by the fact that `D S T` is `t` in the code, and `t = exprType e`, thus of kind `KType`.)

Thus we need to:
- check that `t` is of the form "a type constructor applied to a bunch of things"
- extract the head `D`
- check whether `vc` is a constructor of `D`

I.e. something like (off the top of my head):
`Just tc == fmap fst (decomposeTAppCon t)`

brprice: This is implemented in https://github.com/hackworthltd/primer/pull/1071

georgefst:
Superseded by https://github.com/hackworthltd/primer/pull/1071.

PR #1070, opened 2023-06-10
Rebase of #1040, do not merge to main

dhess:
This is just a rebase of #1040 on current main, so that I can deploy it. (#1040 lacks the necessary GitHub workflow to do this, and I didn't want to touch that PR.)

PR #1071, opened 2023-06-11
fix: better api annotation of valcons matching selection's type

brprice:


brprice:
This implements https://github.com/hackworthltd/primer/pull/1067#discussion_r1225343753

georgefst (approved):
Thank you!

dhess:
Thanks! However, keep in mind that what's currently deployed is not `main` but #1070, so maybe add this to that PR so we can get it online and running?

dhess:
I will rebase #1070 when this PR is merged to main, so that we can get this deployed. Thanks all!

PR #910, opened 2023-03-13
chore(nix): Bump to haskell.nix experimental GHC 9.6 branch.

dhess:

PR #1077, opened 2023-06-18
chore: GHC 9.6.2

dhess:
Upstream blockers:

* https://github.com/ndmitchell/hlint/issues/1477

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -19,6 +19,9 @@ package primer
 package primer-service
   test-options: "--size-cutoff=32768"
 
+-- Temporary during GHC 9.6 transition.
+allow-newer: *:*
+
 -- We need a newer version of Selda than what's been released to Hackage.
 source-repository-package
   type: git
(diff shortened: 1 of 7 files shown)


brprice:
hlint 3.6 is now available in haskell.nix, so this update should probably work once our flake deps have been bumped.

How does this pr relate to #910? (I'm guessing it supersedes it, but https://github.com/hackworthltd/primer/pull/910#issuecomment-1476686738 will probably remain -- this has been fixed upstream, but needs backporting to 9.6 and then releasing.)

brprice:
One other potential ghc 9.6 problem:

We are pinning hedgehog to 1.1, due to https://github.com/hedgehogqa/haskell-hedgehog/issues/487#event-9974826901. This is now fixed and should be released in 1.4 shortly. Note that ghc 9.6 support was introduced in 1.3, so we have a choice between ghc9.4+hedgehog-that-replays and ghc9.6+hedgehog-that-has-replay-bug. However, https://github.com/hedgehogqa/haskell-hedgehog/pull/489#issuecomment-1660039472 suggests a new release should be less than a week away

dhess:
Closing in favor of #1113.

PR #1079, opened 2023-06-19
chore: enable weeder, remove a weed

brprice:
Now that cborg (which is a dependency of weeder) works again for aarch64 (https://github.com/well-typed/cborg/pull/307), we can reenable this check.

We also remove one weed which has grown in the period when weeder was disabled.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -264,7 +264,7 @@
           // primerFlake.packages;
 
           checks = {
-            inherit openapi-validate;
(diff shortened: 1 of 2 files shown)


brprice:
https://buildkite.com/hackworthltd/primer/builds/3589 is a build of this PR + restoring `enhole`, showing that weeder is active in CI again

dhess (approved):

PR #1080, opened 2023-06-20
refactor: move eval into a separate `primer-eval` package

dhess:
Still very much a WIP. The tests need fixing up, and `primer-service` will need to combine the sessions-and-actions API with the factored-out eval API. None of that work has been done yet, but `primer` and `primer-eval` do compile, at least.

I'm open to suggestions/techniques for how best to combine the two APIs.

(One could also imagine a further evolution where we move actions & type checking to the frontend, as well, leaving only the sessions API, i.e., program store, for `primer-service`.)


dhess:
After chatting with @brprice today, I think this is probably the wrong approach, or at least, the wrong order in which to do things. The tentative plan now is:

1. "chop off" the `Primer.API` module and its dependencies into a new `primer-api` package.
2. Split out the eval-related bits of the remaining core package into `primer-eval`, similar to what I've done here.
3. Optionally split out the type checking & editing bits of the core package into, e.g., `primer-typecheck`, leaving `primer-core` as the bits that both `primer-eval` and `primer-typecheck` depend on.

Then we'd have a pretty good story for compiling `primer-eval` to Wasm and running that in the browser, with a minimal footprint of code.

PR #1086, opened 2023-06-28
fix: an eval&undo/redo "lost ID" bug

brprice:


primer/src/Primer/API.hs
```diff
@@ -1109,23 +1109,23 @@ evalFull' ::
...
     q ::
       Maybe TerminationBound ->
       GVarName ->
-      EditAppM (PureLog (WithSeverity l)) Void EvalFullResp
+      QueryAppM (PureLog (WithSeverity l)) Void EvalFullResp
     q lim d = do
-      e <- DSL.gvar d
-      a <- get
+      -- We don't care about uniqueness of this ID, and we do not want to
+      -- disturb any FreshID state, since that could break undo/redo.
```
georgefst, line 1120: It's not _immediately_ obvious why this is safe. I guess because the first eval step will always be to inline the definition, thus removing the ID? Anyway, this is probably worth a comment.

georgefst (approved):
Nice!

I seem to remember you saying earlier this week that you'd realised _two_ issues that could lead to undo bugs (aside from version incompatibility). Did that refer to the last two commits here as separate issues, or is there a further fix needed?

brprice:
> I seem to remember you saying earlier this week that you'd realised _two_ issues that could lead to undo bugs (aside from version incompatibility). Did that refer to the last two commits here as separate issues, or is there a further fix needed?

The two issues were (a) the first commit and (b) the last two commits. The first (as mentioned in commit message) turned out not to be an issue in the deployed version (which explains why I could not manage to get a failure because of it). I am not aware of any other issues causing undo bugs after this PR.

PR #1087, opened 2023-06-30
refactor: move the Primer API into its own package

dhess:
We would like to run evaluations in the browser, using GHC 9.6's support for Wasm and JavaScript targets. We won't need the API wrapper for that, nor type-checking, so this is the first step towards a more fine-grained set of packages in order to reduce the amount of code we need to compile for a Wasm/JavaScript target.

This is a pure refactor. No tests or functionality have been removed, nor any new functionality or tests added. We've simply shuffled things around a bit, and moved some additional code into `primer-testlib` so that it can be reused in `primer-api`'s tests.


dhess:
Assuming there are no major objections to, or omissions in, this PR, could someone please add it to the merge queue once it's been reviewed and accepted?

primer/src/Primer/JSON.hs
```diff
@@ -15,7 +15,7 @@ import Data.Aeson (
  )
 import Deriving.Aeson (CustomJSON (..), OmitNothingFields)
 
--- | A type for Primer API JSON encodings.
+-- | A type for Primer JSON encodings.x
```
brprice, line 18: Typo

primer-api/primer-api.cabal
```diff
@@ -0,0 +1,199 @@
...
+    LambdaCase
+    OverloadedStrings
+
+  ghc-options:
+    -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
+    -Wcompat -Widentities -Wredundant-constraints
+    -Wmissing-deriving-strategies -fhide-source-paths
+
+  build-depends:
+    , aeson                        >=2.0      && <2.2
```
brprice, line 36: I wonder if any build-depends can be trimmed, either here or in the primer package now the api has been split?

brprice (approved):
I have only two minor comments, which I'll fix and then add this to the merge queue

primer-api/primer-api.cabal
dhess: Yes, thanks, I meant to revisit that and forgot before I pushed this PR.

PR #1094, opened 2023-07-11
WIP: more compact pretty printing

brprice:


brprice:
When debugging, I got annoyed at pretty-printed expressions
- taking up so much space
- having superfluous brackets
- printing some things in white (terrible if you use a black-on-white terminal theme)

This PR somewhat addresses all three, but needs some work (see comments in code), and probably splitting into separate PRs

PR #1095, opened 2023-07-13
Provide APIs for viewing and editing kinds of type parameters

georgefst:
We can now finally construct any typedef supported by the backend.

There are a number of things which I intend to punt to https://github.com/hackworthltd/primer/issues/1098. This PR is already big enough.


primer/src/Primer/Action/Available.hs
```diff
@@ -397,13 +403,41 @@ forTypeDefParamNode paramName l Editable tydefs defs tdName td =
...
+      Just (KHole _) -> [NoInput MakeKType]
+      Just _ -> [NoInput DeleteKind]
+  where
+    findKind i k =
+      if getID k == i
+        then Just k
+        else case k of
+          KHole _ -> Nothing
+          KType _ -> Nothing
+          KFun _ k1 k2 -> findKind i k1 <|> findKind i k2
```
georgefst, line 432: This is another thing which can be simplified once we have zippers etc. for kinds.

primer/test/Tests/Action/Available.hs
```diff
@@ -333,14 +334,36 @@ tasty_available_actions_accepted = withTests 500 $
...
+                                )
                               ,
-                                ( typeDefSel $ Just $ TypeDefParamNodeSelection p
-                                , Available.forTypeDefParamNode p l defMut allTypes' allDefs' defName def'
+                                ( 3
+                                , do
+                                    let allKindIDs = \case
+                                          KHole m -> [getID m]
+                                          KType m -> [getID m]
+                                          KFun m k1 k2 -> [getID m] <> allKindIDs k1 <> allKindIDs k2
```
georgefst, line 356: This is another thing which can be simplified once we have zippers etc. for kinds.

primer/src/Primer/App/Base.hs
```diff
@@ -70,7 +72,7 @@ data NodeType = BodyNode | SigNode
 -- | Describes which element of a (type or term) definition the student has selected.
 -- We have the following invariant: when this contains a `NodeSelection` with @nodeType = SigNode@,
 -- or any `TypeDefConsFieldSelection`, then they will always have @meta = Right _@.
-type Selection = Selection' (Either ExprMeta TypeMeta)
+type Selection = Selection' (Either ExprMeta (Either TypeMeta KindMeta))
```
georgefst, line 75: This nested `Either` is maybe a bit ugly, but it's still in some ways simpler than defining a new type, whilst still being fairly self-explanatory.

georgefst:
Sure. We can now construct types like `Fix` or `MaybeT` but we're limited in being able to use them since all foralls implicitly quantify over variables with the kind "type". For example, we can't remove the mismatch here:

image

primer/src/Primer/Core/Type.hs
```diff
@@ -44,7 +47,7 @@ data Type' a
   | TFun a (Type' a) (Type' a)
   | TVar a TyVarName
   | TApp a (Type' a) (Type' a)
-  | TForall a TyVarName Kind (Type' a)
+  | TForall a TyVarName (Kind' ()) (Type' a)
```
brprice, line 50: For clarity: we intentionally don't put metadata on the kind here as is out of scope for this PR. See FR #1098. Here we are only interested in type parameters, and not foralls

primer/src/Primer/Module.hs
```diff
@@ -127,22 +130,26 @@ nextModuleID m = foldl' (\id_ d -> max (nextID d) id_) minBound (moduleDefs m)
...
-  Module
-    { moduleName = primitiveModuleName
-    , moduleTypes = TypeDefPrim <$> M.mapKeys baseName allPrimTypeDefs
-    , moduleDefs = M.fromList $ [(primDefName def, DefPrim def) | def <- enumerate]
-    }
+primitiveModule :: MonadFresh ID m => m Module
+primitiveModule = do
+  allPrimTypeDefs' <- traverse (traverseOf (#primTypeDefParameters % traversed % _2) generateKindIDs) allPrimTypeDefs
+  -- allPrimTypeDefs' <- traverse _ allPrimTypeDefs
+  -- boolDef' <- traverseOf _typedefFields generateTypeIDs $ TypeDefAST boolDef
```
brprice, line 137: Did you mean to commit these comments?

primer-api/src/Primer/API.hs
```diff
@@ -1334,11 +1342,11 @@ getSelectionTypeOrKind = curry $ logAPI (noError GetTypeOrKind) $ \(sid, sel0) -
...
       case sel.node of
         -- type def itself selected - return its kind
-        Nothing -> pure $ Kind $ viewTreeKind $ typeDefKind $ TypeDef.TypeDefAST def
+        Nothing -> pure $ Kind $ viewTreeKind' $ mkIdsK $ typeDefKind $ forgetTypeDefMetadata $ TypeDef.TypeDefAST def
         -- param node selected - return its kind
         Just (TypeDefParamNodeSelection p) ->
-          maybe (throw' $ ParamNotFound p) (pure . Kind . viewTreeKind . snd) $
-            find ((== p) . fst) (astTypeDefParameters def)
+          maybe (throw' $ ParamNotFound p.param) (pure . Kind . viewTreeKind . snd) $
+            find ((== p.param) . fst) (astTypeDefParameters def)
```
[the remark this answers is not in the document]
brprice: I'm not a fan of the current behaviour. I think it would be better to (as you suggest) just say "kind".

primer/src/Primer/Action.hs
```diff
@@ -471,6 +471,8 @@ applyAction' a = case a of
   RenameCaseBinding x -> \case
     InBind (BindCase z) -> InBind . BindCase <$> renameCaseBinding x z
     _ -> throwError $ CustomFailure a "cannot rename this node - not a case binding"
+  ConstructKType -> const $ throwError $ CustomFailure ConstructKType "type edits currently only allowed in typedefs"
+  ConstructKFun -> const $ throwError $ CustomFailure ConstructKFun "type edits currently only allowed in typedefs"
```
brprice, line 475: Should these errors read "kind edits"?

primer/src/Primer/App.hs
```diff
@@ -983,11 +984,49 @@ applyProgAction prog = \case
                                 Just
                                   TypeDefConsFieldSelection
                                     { index
-                                    , meta = Right $ zt ^. _target % _typeMetaLens
+                                    , meta = Right $ Left $ zt ^. _target % _typeMetaLens
                                     }
                             }
                   }
           )
+  ParamKindAction tyName paramName id actions -> editModuleOfCrossType (Just tyName) prog $ \(mod, mods) defName def -> do
```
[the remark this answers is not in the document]
brprice: > The current approach is not ideal since we traverse both branches even though only one will have a matching ID, and we don't fail if the ID isn't found anywhere

Do you think it is worth fixing this in this PR? I think it would be somewhat easy to do, but perhaps not worth it if you expect to revisit `Loc` soon.
Three ways spring to mind, all roughly implementing a `modifyKindNode :: ID -> Kind -> (Kind -> Kind) -> Either Error Kind` (or some variety of `focusKind`) function (where error could be "id not found" or "multiple id"):
- just do it in the "findNode" style -- we may already have the ability to use a generic uniplate zipper on things which `HasID` to find a particular node
- since kinds are simple, we could use some uniplate magic: something like `contexts` to get a focus on each node, then `filter` this by ID, then do the transform
- some small applicative-state hack on this current code where we work not in the current monad `m`, but in something like `StateT Int m` (or maybe `AccumT` or `WriterT` may be better) where every time the `modifyKind` actually runs the function, it also emits a `+1`. We then check that we counted to exactly `1` (i.e. did exactly one edit). (It may be nicer to work in the "0-1-many" monoid, rather than the integers, as that more clearly expresses what we care about).

> We also end up performing a full typechecking pass

This seems fine to me (at least, it is consistent with other actions)

primer-service/test/outputs/OpenAPI/openapi.json
```diff
@@ -318,7 +318,10 @@
                     "DeleteCon",
                     "AddConField",
                     "DeleteConField",
-                    "DeleteTypeParam"
+                    "DeleteTypeParam",
+                    "MakeKType",
+                    "MakeKFun",
+                    "DeleteKind"
```
brprice, line 324: This commit should maybe be a breaking change? We may emit new stuff from the api

primer/src/Primer/Action.hs
```diff
@@ -1290,7 +1309,7 @@ toProgActionInput def0 sel0 opt0 = \case
     opt <- optNoCxt
     sel <- typeSel
     index <- length . astTypeDefParameters <$> typeDef -- for now, we always add on to the end
-    pure [AddTypeParam sel.def index opt C.KType]
+    pure [AddTypeParam sel.def index opt $ C.KHole ()]
```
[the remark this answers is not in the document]
brprice: Good question. I'm not sure.

There is one sense in which kinds are different to terms/types: since they are so simple, there is a canonical thing to put in the leaves of a non-holey tree: a `KType`. Since our kind actions will all push the current kind down (i.e. `k ~> k -> ?`), one will always end up filling in this hole with a `KType`. (This is not true in types/terms, since there are more options for leaves.).

Indeed, if we are going the direction of "leave as `KType` because of the above justification", I would perhaps argue that inserting an arrow should not create `k -> ?`, but `k -> KType`.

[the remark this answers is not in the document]
georgefst: Interesting. One could argue that we should go a step further and not allow constructing `KHole` at all. In practice I suppose this would mean, in addition to reverting the change above, replacing the "delete kind" action with a "raise" action, offered only on `KFun` nodes.

Now that I think about it, the purpose of `KHole` is somewhat unclear, given that we have such a simple kind system. Is it just used to give a kind to `TEmptyHole`? And would it be reasonable to just say that that has the kind `KType` instead?

primer-api/src/Primer/API.hs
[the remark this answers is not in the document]
georgefst: It would be easy enough to implement, changing `data TypeOrKind = Type Tree | Kind Tree` to `data TypeOrKind = Type Tree | Kind Tree | SortKind` (naming?). But I don't know how we would want to communicate this last case in our `primer-app` UI.

@dhess Any thoughts?

primer/src/Primer/Core/Type.hs
georgefst: Since I already need to rebase, I'll add this to the commit message when doing so.

primer/src/Primer/Action.hs
georgefst: Yes, will edit.

primer-service/test/outputs/OpenAPI/openapi.json
georgefst: Yes, you're right.

primer-api/src/Primer/API.hs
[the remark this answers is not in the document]
dhess: I think I understand the question, but let's make sure. Is this about what type (aka "kind") to display when a kind is, say, `* -> * -> *`, and the student selects a non-root node in that tree?

primer/src/Primer/Action.hs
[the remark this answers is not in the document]
brprice: > Interesting. One could argue that we should go a step further and not allow constructing `KHole` at all. In practice I suppose this would mean, in addition to reverting the change above, replacing the "delete kind" action with a "raise" action, offered only on `KFun` nodes.

Potentially, but see below about why `KHole` exists (and is not easily going away). Thus if we forbade constructing `KHole` at all, it would be somewhat inconsistent.

Being able to give hole kinds does enable students to write more "programs" (i.e. things the system accepts, but are obvious nonsense). This may be good or bad depending on whether this is the sort of confusion/freedom that is good for learning(!), and on whether these states actually have utility as intermediate states in interactive use (I have not got any intuition on this as yet). In particular, one could write something like `data T (p : ?) = C (p p) p`, where `p` is being used at multiple kinds: (roughly speaking) `k -> *`, `k` and `*` (for some unknown `k`). It is not possible to fill in the `KHole` here to get a valid complete program. (As an aside, it is possible to make sense of this in a richer type system, by giving `p` a polymorphic kind; e.g. ghc will accept `data D (p :: forall k . k) = C p (p p)`. I don't know whether things of this ilk are practically useful.)

> Now that I think about it, the purpose of `KHole` is somewhat unclear, given that we have such a simple kind system. Is it just used to give a kind to `TEmptyHole`? And would it be reasonable to just say that that has the kind `KType` instead?

It is indeed used to kind `TEmptyHole` (iirc). In the past we did say `TEmptyHole` has kind `KType`, but there was an issue with that, and I changed it https://github.com/hackworthltd/vonnegut/commit/0f291a3db0904e1d05b1af60201358565089ac9a . In particular, we really want to be able to create type applications via the sequence (e.g.) `?  ;   ? ?   ;   List ?  ;  List Bool`, but the intermediate type `? ?` is checked by synthesising a kind for the first hole, ensuring it is (consistent with) an arrow kind and the second hole matches the domain -- this would fail if we gave the kind `KType`. (This is not the only way to solve this problem, but is the one we currently use, and would need to revisit this if we wanted to give the kind `KType` to type holes.)

primer-api/src/Primer/API.hs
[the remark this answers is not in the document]
georgefst: Prior to this PR, we have a function `Selection -> (Type | Kind)`. When any value-level node is selected it returns its type, and when any type-level node is selected we return its kind. The question is essentially: now that a selection can be of a kind-level node, what should be returned in that case?

(which I think is almost what you were saying, except that it includes root nodes)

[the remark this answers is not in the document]
dhess: Let's discuss in the upcoming developer meeting.

PR #1109, opened 2023-08-10
fix: `maxID` looks at type ids

brprice:
Previously it would miss ids of types in annotations (or type applications).


georgefst (approved):

brprice:
Thought I should add a test (especially as we may be needing to add kind ids soon!), resubmitting for review.

PR #1113, opened 2023-08-16
chore: bump to GHC 9.6.2

dhess:


brprice:
Note that https://github.com/hackworthltd/primer/pull/910#issuecomment-1476686738 is still an issue.

Also, does this supersede #910 and #1077 ?

brprice:
FYI, I have now opened #1117 which should fix any hedgehog-related issues (i.e. those around `Gen.element` and ambiguous Traversables)

dhess:
OK! The project now builds fine with GHC 9.6.2. I think I'll temporarily disable Weeder so that it's not in the critical path for Wasm builds.

PR #1117, opened 2023-08-16
chore: upgrade Hedgehog to version 1.4

brprice:
Now that Hedgehog 1.4 has been released with [the fix for replaying tests](https://github.com/hedgehogqa/haskell-hedgehog/pull/489), we can migrate off 1.1. This is a prerequisite for upgrading ghc to 9.6 (#1077).

Note that currently we need to `allow-newer` to get hedgehog-classes to work with hedgehog 1.4; I have opened an upstream PR at https://github.com/hedgehogqa/haskell-hedgehog-classes/pull/52

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -9,6 +9,8 @@ packages:
 
 optimization: 0
 
+allow-newer: hedgehog-classes:hedgehog
+
 package *
   ghc-options: -fwrite-ide-info
   benchmarks: True
--- a/primer-api/primer-api.cabal
+++ b/primer-api/primer-api.cabal
@@ -67,7 +67,7 @@ library primer-api-hedgehog
 
   build-depends:
     , base
-    , hedgehog                                          ^>=1.1
+    , hedgehog                                          ^>=1.4
     , primer-api
     , primer:{primer, primer-hedgehog, primer-testlib}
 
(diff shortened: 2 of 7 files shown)


cabal.project
```diff
@@ -9,6 +9,8 @@ packages:
 
 optimization: 0
 
+allow-newer: hedgehog-classes:hedgehog
+
```
brprice, line 13: I intentionally did not sign-off this commit so CI would break and this would not be merged without discussing:
- Do we want to merge this with `allow-newer` (if so, I need to reword the commit message to be conventional-commits, and maybe rebase it first so both commits will build cleanly; I'll not squash, to make it trivial to revert once upstream makes a release)
- Do we want to wait for upstream to fix it (https://github.com/hedgehogqa/haskell-hedgehog-classes/pull/52)
- Did we want some other option?

dhess: I think this is fine. It's more important that we get the GHC 9.6 work moving.

dhess (approved):
Thanks!

cabal.project
brprice: Upstream plans to cut a new release this weekend, so I'll wait until next week to merge this.

brprice: Let's merge. I have opened #1122 to track removing the `allow-newer` when a release happens.

PR #1120, opened 2023-08-17
feat: display and edit kinds on foralls

brprice:
Notes for reviewers: Sorry the PR is so large; I think most of it is fairly mechanical refactors and I have tried to split those into separate commits where I can


primer/src/Primer/TypeDef.hs
```diff
@@ -106,26 +105,21 @@ typeDefAST = \case
...
-forgetTypeDefMetadata =
-  over _typedefFields forgetTypeMetadata
-    . over _typedefParamKinds forgetKindMetadata
+forgetTypeDefMetadata (TypeDefPrim td) = TypeDefPrim $ over (#primTypeDefParameters % mapped % _2) forgetKindMetadata td
+forgetTypeDefMetadata (TypeDefAST (ASTTypeDef ps cs hs)) =
+  TypeDefAST $
+    ASTTypeDef
+      (over (mapped % _2) forgetKindMetadata ps)
+      (over (mapped % #valConArgs % mapped) forgetTypeMetadata cs)
+      hs
```
brprice, line 118: It is unfortunate that this has got more verbose. I cannot see how to do this nice and lens-y (compare with old code above) because we now cannot change metadata in params independently to in ctors, as both controlled by same `c`. (The same comment applies to `generateTypeDefIDs`)

primer/src/Primer/TypeDef.hs
```diff
@@ -105,6 +105,9 @@ typeDefAST = \case
 typeDefKind :: TypeDef b () -> Kind' ()
 typeDefKind = foldr (KFun () . snd) (KType ()) . typeDefParameters
 
+-- TODO/REVIEW: I cannot see how to do this nice and lens-y (see old code
+-- above) because we cannot change metadata in params indep to in ctors,
+-- as both controlled by same `c`
```
brprice, line 110: TODO: I'll rebase this commit away after reviews are done (it has been migrated to a review comment)

primer/src/Primer/Action/Available.hs
```diff
@@ -167,7 +167,8 @@ data NoInputAction
   | DeleteTypeParam
   | MakeKType
   | MakeKFun
-  | DeleteKind
+  | -- TODO: RaiseKind?
```
georgefst, line 170: I assume you meant to leave a review comment here?

Anyway, yes, we probably should have a "raise kind" action. Though I don't mind whether that's in a separate PR.

primer/src/Primer/Typecheck.hs
```diff
@@ -814,6 +814,10 @@ check t = \case
             Hole{} -> default_ -- Don't let the recursive call mint a hole.
             e'' -> pure e''
       (Hole _ (Ann _ _ ty), SmartHoles)
+        -- TODO/REVIEW: when testing editable-forall-kinds, I noticed that we would not elide the hole in
+        --   foo : ∀a:*.? ; foo = {? ? : ∀a:*.? ?}
+        -- but this is odd -- the annotations are identical (so not acting as a typechange)!
+        -- (also, note that we keep hole-wrapped-holey-anns around because of the below even when we would accept the non-wrapped version (indeed, have an action that will do that!)
```
georgefst, line 820: This is odd. DO you have any inkling as to why it occurs?

primer/src/Primer/TypeDef.hs
georgefst: Is it possible to at least write a lens which is as general as possible in updating kinds in all positions, and use that here?

georgefst (approved):
Great stuff. Thank you for taking on this work!

primer-api/src/Primer/API.hs
```diff
@@ -750,7 +751,7 @@ viewProg p =
                           astTypeDefConstructors t <&> \(TypeDef.ValCon nameCon argsCon) ->
                             ValCon
                               { name = nameCon
-                              , fields = viewTreeType' . over _typeMeta (show . view _id) <$> argsCon
+                              , fields = viewTreeType' . over _typeKindMeta (const @_ @() "") . over _typeMeta (show . view _id) <$> argsCon
```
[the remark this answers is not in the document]
georgefst: OK, I think this is addressed.

primer/src/Primer/Action/Available.hs
georgefst: See also https://github.com/hackworthltd/primer/issues/1141#issue-1892600830.

primer/src/Primer/TypeDef.hs
brprice: I did wonder about that, and decided that it would be a one-off slightly odd definition and not worth it. However, your comment made me realise that we can just use generic optics, which I think is actually pretty nice:
```haskell
forgetTypeDefMetadata :: TypeDef b c -> TypeDef () ()
forgetTypeDefMetadata =
  set (param @1) ()
    . set (param @0) ()

generateTypeDefIDs :: MonadFresh ID m => TypeDef () () -> m (TypeDef TypeMeta KindMeta)
generateTypeDefIDs =
  traverseOf (param @1) (\() -> meta)
    <=< traverseOf (param @0) (\() -> kmeta)
```

georgefst: Aha, yeah, of course. Nice!

primer/src/Primer/Action/Available.hs
brprice: This is now tracked in #1145

primer/src/Primer/Typecheck.hs
brprice: Yes. This is actually #7. The current design is
- When synthesising "never elide holes", i.e. do not automatically do `{? e ?} ~> e` when we have no information about the context. NB: if we did elide, it could change the synthesised type and force new holes to appear elsewhere in the program.
- When checking `T ∋ {? e ?}` we attempt to elide the hole if that would be type-correct by the transformations
  - if checking nested holes `T ∋ {? {? e ?} ?}`, drop one layer and recurse `T ∋ {? e ?}`
  - if have an annotation in a hole (this is the case referred to in this comment) `T ∋ {? s : S ?}`, we
    - drop a hole-and-trivial-annotation, if it is still type-correct `{? s : ? ?}` changes to `s`, if we have `T ∋ s`
    - never touch holes-and-incomplete-annotations (this is our current example), i.e. if there is a type hole in the annotation `S`, as (quoting Typecheck.hs) 
      > Don't want to, e.g., remove {? λx.x : ? ?} to get λx.x : ? Since holey annotations behave like non-empty holes, we will not elide non-empty holes if they have a holey annotation. (This is needed for idempotency, since we return non-empty holes with holey-annotated contents in the case a construction cannot typecheck, e.g. Bool ∋ λx.t returns {? λx.t : ? ?}
    - otherwise fall through to the "have a hole, and don't care about the contents" case
  - if none of the above are true, we have a hole and don't care about the contents, in which case we elide the hole leaving `e` as long as that is type-correct

(NB: above, where I say "elide...as long as that is type correct" and similar, by "type correct" I actually mean "smart holes accepts the term and does not create a hole _at top level_", i.e. it could result in "pushing down the hole")

dhess: Can you add a comment in the code here with a link to the parent post, and also a link to #7, please?

brprice: > Can you add a comment in the code here with a link to the parent post, and also a link to #7, please?

Will do, but in a separate PR since it is orthogonal to the current changes.

PR #1129, opened 2023-09-04
fix: tcWholeProg updates metadata of ADT parameter selections

brprice:


brprice:
The first commit addresses https://github.com/hackworthltd/primer/pull/1095#discussion_r1275523570, and the second commit fixes (what I think is a latent) a bug where we don't update some selection's metadata. (It is latent since the only consumer is the openapi, and there we always just report `KType` in the API when a node in the kind of a typedef's parameter is selected. I can't recall off the top of my head whether this data is exposed in the "rich" api as well.)

@georgefst Since you implemented #1095, could you give this extra scrutiny (I'm not sure that the new behaviour is now correct -- the story around selections etc is slightly confusing to me)

Context: this code needs changing for #1120 anyway, and I noticed that it seemed weird currently.

primer/src/Primer/Zipper/Type.hs
```diff
@@ -58,11 +59,14 @@ import Primer.Core.Meta (
...
 import Primer.Core.Type (
+  Kind',
   Type' (TForall, TLet),
   TypeMeta,
  )
 import Primer.Name (Name)
 
+type KindZip' c = Zipper (Kind' c) (Kind' c)
+
 type TypeZip' b = Zipper (Type' b) (Type' b)
```
brprice, line 71: I intentionally do not define `KindZip` since it will not be used, and I want to play a small trick with it in #1120 to split my commits nicer

dhess:
Leaving this one for @georgefst.

primer/src/Primer/Core/Type.hs
```diff
@@ -91,6 +92,12 @@ data Kind' a
 _kindMeta :: Traversal (Kind' a) (Kind' b) a b
 _kindMeta = param @0
 
+-- | A lens on to the metadata of a kind.
+-- Note that unlike '_kindMeta', this is shallow i.e. it does not recurse in to sub-expressions.
+-- And for this reason, it cannot be kind-changing.
+_kindMetaLens :: Lens' (Kind' a) a
```
brprice, line 98: This should probably say "... cannot be **type**-changing"

georgefst (approved):
> I'm not sure that the new behaviour is now correct

Yes, this looks correct to me!

> the story around selections etc is slightly confusing to me

Is there anything in particular I can clarify?

primer/src/Primer/Zipper/Type.hs
brprice: FTR, I'm not sure I'm actually going to do any tricks here, but there is no harm in not defining it until #1120

brprice:
> > I'm not sure that the new behaviour is now correct
> 
> Yes, this looks correct to me!

Thanks

> > the story around selections etc is slightly confusing to me
> 
> Is there anything in particular I can clarify?

I haven't currently got any precise questions, but next time I am working in this area I'll distill my confusion and ping you about it.

PR #1134, opened 2023-09-06
fix: nextProgID cares about ids in typedefs

brprice:


brprice:
This has been a latent bug for a long time, and it has just bitten me by making some test outputs very confusing (some actions were creating IDs that seemed far too low)

dhess (approved):

brprice:
Huh, this doesn't actually solve my problem: `unit_ParamKindAction_2` is still creating nodes with small ids: (4 & 5, iirc, where the nextID is ~30)

brprice:
I have added another commit so `ParamKindAction` will error if an ID is not found. This enforces that the now bogus `5` in the test is updated.

primer/src/Primer/App.hs
```diff
@@ -1025,13 +1028,16 @@ applyProgAction prog = \case
         >>= either (throwError . ActionError) pure
     pure (mods', Nothing)
     where
-      modifyKind f k =
+      modifyKind f k
+        | notElemOf (_kindMeta % _id) id k = throwError' $ IDNotFound id
+        | otherwise = modifyKind' f k
```
brprice: FTR, The commit messages says
> The implementation is rather crude, and will be improved shortly when we create a zipper for kinds and use it to replace `modifyKind` and `findKind`.

This is done in #1129

primer/src/Primer/Def/Utils.hs
```diff
@@ -26,6 +27,19 @@ nextID (DefAST (ASTDef e t)) =
...
+-- Note: do not rely on the implementation of this function, as it may
+-- change in the future.
+nextIDTypeDef :: TypeDef TypeMeta KindMeta -> ID
+nextIDTypeDef (TypeDefAST (ASTTypeDef ps vcs _)) =
+  let pid = foldlOf' (folded % _2 % kindIDs) max minBound ps
+      vcid = foldlOf' (folded % #valConArgs % folded % typeIDs) max minBound vcs
+   in succ $ max pid vcid
+nextIDTypeDef (TypeDefPrim (PrimTypeDef ps _)) =
+  succ $ foldlOf' (folded % _2 % kindIDs) max minBound ps
+{-# INLINE nextIDTypeDef #-}
```
georgefst, line 41: I'm intrigued, given that we have never really done any serious low-level optimising, about why we always want to inline this particular? This and the function above are the only two such pragmas in the codebase.

georgefst (approved):

primer/src/Primer/Def/Utils.hs
dhess: I don't recall doing that. 

I don't mind removing it so long as it doesn't adversely affect the benchmarks, but on the other hand, I doubt it's dong any harm, either.

PR #1151, opened 2023-09-26
test: Primitive functions can exhibit laziness

### LINKED ISSUE #1118: Add a test to ensure that primitive functions can be lazily evaluated
We recently re-opened discussions on https://github.com/hackworthltd/primer/issues/272, and decided that the current implementation of primitive functions is sufficient. However, we do want to ensure that a clever primitive function can lazily evaluate its arguments; e.g., it should be possible to write a primitive function like `primitiveConst True ?` and have it return `True` without getting stuck on the hole.

Let's add a test(s) to the test suite to verify these semantics.

georgefst:
Closes #1118.

Note that while this test captures the key property (that primitives can be lazy in an argument), we haven't implemented all suggestions in https://github.com/hackworthltd/primer/issues/1118#issuecomment-1689685996, for reasons outlined in the commit message.

### DIFF
--- a/primer/src/Primer/Primitives.hs
+++ b/primer/src/Primer/Primitives.hs
@@ -49,6 +49,7 @@ import Primer.Core.DSL (
   int,
   tcon,
  )
+import Primer.Core.Utils (generateIDs)
 import Primer.JSON (CustomJSON (..), PrimerJSON)
 import Primer.Name (Name)
 import Primer.Primitives.PrimDef (PrimDef (..))
@@ -135,6 +136,7 @@ primDefName = \case
   IntNeq -> "Int.≠"
   IntToNat -> "Int.toNat"
   IntFromNat -> "Int.fromNat"
+  PrimConst -> "const"
 
 primDefType :: PrimDef -> Type' () ()
 primDefType = uncurry (flip $ foldr $ TFun ()) . primFunTypes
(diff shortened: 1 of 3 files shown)


primer/test/Tests/EvalFull.hs
```diff
@@ -1692,6 +1693,20 @@ unit_prim_ann =
...
         s <~==> Right r
 
+unit_prim_lazy :: Assertion
+unit_prim_lazy =
+  let ((e, r, prims), maxID) =
+        create $
+          (,,)
+            <$> pfun PrimConst
+            `app` bool_ True
+            `app` emptyHole
```
brprice, line 1703: Could we also have a test that `const` does not force its second argument? i.e. with something like `letrec x = x : Nat in x` in place of this `emptyHole`?

georgefst: Sure, now added (I was thinking it didn't really add much over the first test, but I guess it does make sure we don't eagerly go off and prioritise expanding `let`s).

primer/src/Primer/Primitives.hs
```diff
@@ -270,6 +275,10 @@ primFunDef def args = case def of
     [exprToNat -> Just n] ->
       Right $ int $ fromIntegral n
     _ -> err
+  PrimConst -> case args of
+    [x, _] ->
+      Right $ generateIDs x
+    _ -> err
```
brprice, line 281: (Note from after this was merged)

It turns out that this definition is slightly incorrect, for bidirectional reasons. See #1159

PR #1153, opened 2023-09-28
Allow typedef deletions regardless of whether type is in use 

georgefst:
Closes https://github.com/hackworthltd/primer/issues/401.


dhess:
What's the status of this PR? How much work remains before it's ready for review?

primer/src/Primer/App.hs
```diff
@@ -824,24 +849,36 @@ applyProgAction prog = \case
...
+                    $ bs
+                    & filter \case
+                      CaseBranch (PatCon vcName') _ _ | vcName' == vcName -> False
+                      _ -> True
+              )
+              <=< transformM \case
+                Con _ c _ | c == vcName -> emptyHole
+                e -> pure e
+          )
+    ms'' <- runFullTCPass (progSmartHoles prog) (progImports prog) ms'
```
brprice, line 880: You say

> ### Deleting a constructor
>  * Replace all uses of the constructor with a hole
>  * Remove the corresponding branch from any pattern matches 
>  * Run a full typechecking pass
>     * In this instance, I haven't worked out precisely why it is necessary - I just know that without it, the "available actions accepted" property test fails quickly on some massive counterexample

This is somewhat surprising to me (though not shocking), and I thought I'd have a quick look to see if the cause is obvious. However, when I change this line to `let ms'' = ms'`, I don't see any errors running `cabal run primer-test -- -p 'available actions accepted'`. Is this typechecking pass what you meant? Could the problem have been something else what you fixed during development?

primer/src/Primer/App.hs
```diff
@@ -692,14 +692,39 @@ applyProgAction prog = \case
...
-      Just (TypeDefPrim _) -> throwError $ TypeDefIsPrim d
-      Just (TypeDefAST td) -> do
-        checkTypeNotInUse d td $ m : ms
-        let m' = m{moduleTypes = Map.delete (baseName d) (moduleTypes m)}
-        pure (m' : ms, Nothing)
+  DeleteTypeDef d -> editModuleOfCrossType (Just d) prog $ \(m, ms) defName def -> do
+    let updateExpr = \case
+          Con _ c _ | c `elem` map valConName (astTypeDefConstructors def) -> emptyHole
+          e -> pure e
+    ms' <-
```
brprice, line 699: It would be nice to replicate your comments from the PR into the code (here and elsewhere), otherwise future readers will probably lose that context (e.g. why not do anything about pattern matches?).

primer/src/Primer/App.hs
```diff
@@ -692,14 +692,39 @@ applyProgAction prog = \case
...
+                    % traversed
+                )
+                updateType
+          )
+    ms'' <- runFullTCPass (progSmartHoles prog) (progImports prog) ms'
+    pure (ms'', Nothing)
+    where
+      updateType = transformM \case
+        TCon _ n | n == d -> tEmptyHole
+        e -> pure e
```
brprice, line 727: It is somewhat odd to have `updateExpr` defined above in a `let`, but `updateType` defined below in a `where`. Any particular reason?

primer/src/Primer/App.hs
brprice: In fact, adding a `unless (ms'' == ms') $ error "TC changed modules"` line also does not trigger

primer/src/Primer/App.hs
```diff
@@ -894,31 +931,47 @@ applyProgAction prog = \case
...
+              tdName
+              vcName
+              ( \_ (CaseBranch vc binds e) -> do
+                  (binds', Bind _ var) <- deleteIndex binds
+                  CaseBranch vc binds'
+                    <$> ( e & transformM \case
+                            Var _ v | v == LocalVarRef var -> emptyHole
+                            e' -> pure e'
+                        )
+              )
```
brprice, line 958: I'm suspicious of this inner "v ~> ?" transformation, in the presence of shadowing. What would happen when deleting the parameter in `case e of C x -> MkPair x (\x -> x)`, where the last `x` does not refer to the pattern variable? Perhaps `traverse` over `_freeVars` would be more appropriate?

primer/test/Tests/Action/Prog.hs
```diff
@@ -873,6 +873,27 @@ unit_copy_paste_import =
...
         Right assertion -> assertion
 
+unit_DeleteTypeDef :: Assertion
+unit_DeleteTypeDef = progActionTest
+  ( defaultProgEditableTypeDefs
+      $ sequence
+        [ do
+            x <- emptyHole `ann` (tcon tT `tapp` tcon (tcn "Bool") `tapp` tEmptyHole)
+            astDef "def" x <$> tEmptyHole
+        ]
```
brprice, line 883: Might be better to also use the type's constructors at term level, both as constructors and as patterns in a `case`

primer/test/Tests/Action/Prog.hs
```diff
@@ -1076,6 +1097,37 @@ unit_AddCon_sparse =
...
+  ( defaultProgEditableTypeDefs
+      $ sequence
+        [ do
+            x <-
+              case_
+                (emptyHole `ann` (tcon tT `tapp` tcon (tcn "Bool") `tapp` tcon (tcn "Int")))
+                [ branch cA [("x", Nothing), ("y", Nothing), ("z", Nothing)] emptyHole
+                , branch cB [("s", Nothing), ("t", Nothing)] emptyHole
+                ]
+            astDef "def" x <$> tEmptyHole
```
brprice, line 1111: Could we also use `cA` as a constructor? (Maybe in the RHS of the `cB` branch)

primer/test/Tests/Action/Prog.hs
```diff
@@ -1158,6 +1210,52 @@ unit_AddConField_case_ann =
...
+                cA
+                [ con0 (vcn "True")
+                , con0 (vcn "False")
+                , con0 (vcn "True")
+                ]
+                `ann` (tcon tT `tapp` tEmptyHole `tapp` tEmptyHole)
+            )
+            [ branch cA [("x", Nothing), ("y", Nothing), ("z", Nothing)] emptyHole
+            , branch cB [("s", Nothing), ("t", Nothing)] emptyHole
+            ]
```
brprice, line 1229: Could we also use `cA` as a constructor?

brprice (changes requested):
Looks pretty good. I have a handful of small comments and one large one that has triggered me to put "Request changes": the issue about scoping when deleting a field.

primer/src/Primer/App.hs
brprice: Looking briefly at these errors, they seem to (almost?) all be only changes in typecaches, and even only filling in the typecache of the just-created holes. This is completely expected. To validate this, I ran `cabal run primer-test -- -p "available actions accepted"` in a loop with the below patch (remove the TC pass, and relax the test) and got drastically reduced numbers of errors. However, I did still see
- `cabal run primer-test -- -p "available actions accepted" --hedgehog-replay "122/200:aKdB3dEbAhCa3ChA3 Seed 210081999338504562 8735016956953542721"` about `WrongCaseBranches`.
I don't know whether this indicates a problem or not. I would like to at least understand why the procedure "replace uses with hole, delete branches" seems to give malformed `case`s on occasion.

<details>
<summary>The applied patch</summary>

```diff
diff --git a/primer/src/Primer/App.hs b/primer/src/Primer/App.hs
index a6f474ff..20e26868 100644
--- a/primer/src/Primer/App.hs
+++ b/primer/src/Primer/App.hs
@@ -877,7 +877,8 @@ applyProgAction prog = \case
                 Con _ c _ | c == vcName -> emptyHole
                 e -> pure e
           )
-    ms'' <- runFullTCPass (progSmartHoles prog) (progImports prog) ms'
+    --ms'' <- runFullTCPass (progSmartHoles prog) (progImports prog) ms'
+    let ms'' = ms'
     pure (ms'', Just $ SelectionTypeDef $ TypeDefSelection tdName Nothing)
   AddConField type_ con index new ->
     editModuleCross (qualifiedModule type_) prog $ \(m, ms) -> do
diff --git a/primer/test/Tests/Action/Available.hs b/primer/test/Tests/Action/Available.hs
index 278915a0..845c00e5 100644
--- a/primer/test/Tests/Action/Available.hs
+++ b/primer/test/Tests/Action/Available.hs
@@ -141,7 +141,7 @@ import Primer.Gen.Core.Raw (genName)
 import Primer.Gen.Core.Typed (WT, forAllT, genChar, genInt, propertyWT)
 import Primer.Log (PureLog, runPureLog)
 import Primer.Module (
-  Module (Module, moduleDefs),
+  Module (..), --(Module, moduleDefs),
   builtinModule,
   moduleDefsQualified,
   moduleName,
@@ -175,6 +175,7 @@ import Tests.Typecheck (
   runTypecheckTestMIn,
  )
 import Text.Pretty.Simple (pShowNoColor)
+import Primer.App.Utils (forgetProgTypecache)
 
 -- | Comprehensive DSL test.
 test_1 :: TestTree
@@ -502,7 +503,7 @@ tasty_available_actions_accepted = withTests 500
         (_, (Right _, a'')) -> ensureSHNormal a''
     ensureSHNormal a = case checkAppWellFormed a of
       Left err -> annotateShow err >> failure
-      Right a' -> TypeCacheAlpha a === TypeCacheAlpha a'
+      Right a' -> TypeCacheAlpha (forgetProgTypecache $ appProg a) === TypeCacheAlpha (forgetProgTypecache $ appProg a') --TypeCacheAlpha a === TypeCacheAlpha a'
 
 -- 'Raise' works when moving checkable terms into synthesisable position
 unit_raise_sh :: Assertion
```
</details>

brprice: Ah, I understand that issue: we have `data T = C` and `foo = case (? : T) of _ -> ?` (note the wildcard). Since we don't change wildcards according to the above algorithm, we end up with `data T = <empty>`, `foo = case (? : T) of _ -> ?`. But TC complains about this fallback branch, as it expects an empty case `foo = case (? : T) of {}`!

The TC will fix this up fine, but it would be more consistent (with removing named branches) to manually remove such fallbacks.

primer/src/Primer/Action/Available.hs
```diff
@@ -462,9 +460,9 @@ forTypeDefConsFieldNode ::
   ASTTypeDef TypeMeta KindMeta ->
   [Action]
 forTypeDefConsFieldNode _ _ _ _ NonEditable _ _ _ _ = mempty
-forTypeDefConsFieldNode con index id l Editable tydefs defs tdName td =
+forTypeDefConsFieldNode con index id l Editable _ _ _ td =
```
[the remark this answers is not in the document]
georgefst: Maybe. I suspect we may well re-add them at some point, and even with them unused, the uniformity helps in simplifying the definition of `Primer.API.availableActions`. Ideally we'd probably refactor this to pass around a record for the common fields, in which case we'd be unlikely to complain that not all fields of the record are used in every function which receives it.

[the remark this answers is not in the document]
brprice: Fair point. I don't have strong opinions either way.

primer/src/Primer/App.hs
georgefst: Thanks for looking in to this. I've added (a slightly cleaned-up version of) your patch, and edited OP accordingly.

primer/src/Primer/App.hs
georgefst: Agreed, but we mostly use `where` in the other cases, but `updateExpr` can't go there because then `def` would be out of scope. It's a bit weird, but I don't think it's worth changing.

georgefst: Agreed, but we mostly use `where` in the other cases, but `updateExpr` can't go there because then `def` would be out of scope. It's a bit weird, but I don't think it's worth changing.

primer/src/Primer/App.hs
georgefst: Good catch, that's a pretty silly mistake from me! I've now replaced with a `traverseOf _freeTmVars`. I'd have hoped that our property tests might catch this sort of thing, but I guess there's nothing biasing the generators towards creating shadowing scenarios.

georgefst:
> > What's the status of this PR? How much work remains before it's ready for review?
>
> None! I just somehow forgot to ask for a review.

Actually, I think it was because I had intended to add a few more tests. I'm just finishing these now, along with @brprice's suggested tests.

brprice (approved):

primer/src/Primer/App.hs
georgefst: > However, I don't know what test would actually catch this, as it would still be type-correct, just delete more that expected.

Oh, good point.

primer/src/Primer/App.hs
georgefst: Ah, `null $ astTypeDefConstructors def'` isn't actually quite the right check. What if the fallback only covered one constructor, and this action deletes it? Then we'd have a meaningless wildcard. I've changed it to `map (PatCon . valConName) (astTypeDefConstructors def') == map caseBranchName bs'`, where `bs'` is the first element of the tuple.

This was actually caught by the property test when I tried to merge: `cabal run primer-test -- -p "actions accepted" --hedgehog-replay "417/1032:e2Id2J Seed 1333018207506774433 10605815445256691107"`.

PR #1158, opened 2023-10-02
style: fourmolu 0.14.0.0

dhess:
Note that there are no functional changes here; this is pure formatting.


brprice (approved):

georgefst (approved):
For the record, I don't actually like some of these changes, and I really hoped that Ormolu would have stabilised more now than it's been about four years. But we don't have much choice other than to roll with it.

PR #1164, opened 2023-10-17
Animations

georgefst:
This is a proof-of-concept, but the implementation is solid enough to merge. Follow-up work is tracked in #1173.

(note to self - if any significant changes are made here due to reviews, #1173 will probably need to be modified accordingly)


dhess:
Is the intent of this PR to merge it into `main`, or is it just a playground for some ideas?

georgefst:
> Is the goal of this PR to merge it into `main`, or is it just a playground for some ideas?

To merge. This code is now pretty solid, so I see no harm in keeping it around, even if some of it will be rewritten as we change our implementation or the student-facing API.

The one unsatisfying part is that we're extremely limited in what sort of lambdas we can pass to our `animate` primitive (I'll expand on this shortly). This is relaxed by https://github.com/hackworthltd/primer/pull/1169, but that is, unlike this PR now, very much an unmergeable hack.

brprice:
> The one unsatisfying part is that we're extremely limited in what sort of lambdas we can pass to our `animate` primitive (I'll expand on this shortly). This is relaxed by #1169, but that is, unlike this PR now, very much an unmergeable hack.

Could you point to where you expanded on this?

georgefst:
> > The one unsatisfying part is that we're extremely limited in what sort of lambdas we can pass to our `animate` primitive (I'll expand on this shortly). This is relaxed by #1169, but that is, unlike this PR now, very much an unmergeable hack.
> 
> Could you point to where you expanded on this?

Where it says "(full explanation to be filled in imminently)"! Sorry, I'll nudge you when that's done.

primer/src/Primer/Builtins/Picture.hs
```diff
@@ -0,0 +1 @@
+module Primer.Builtins.Picture () where
```
brprice, line 1: Did you mean to commit this empty module? (You have referred to it in `primer.cabal` also). Maybe add a comment explaining the purpose?

primer/src/Primer/Primitives.hs
```diff
@@ -276,6 +333,49 @@ primFunDef def args = case def of
     [exprToNat -> Just n] ->
       Right $ int $ fromIntegral n
     _ -> err
+  Animate -> case args of
+    -- Since we can only translate a `Picture` expression to an image once it is in normal form,
+    -- this guard will only pass when `picture` has no free variables other than `time`.
+    [PrimCon () (PrimInt duration), Lam () time picture]
+      | Just frames <- traverse diagramAtTime [0 .. duration * 100 `div` frameLength] ->
```
brprice: (Pedantic off-by-one): I think this length of frame indices is one too long (not that a human would be able to visually tell unless the construction function was wild)!

brprice:
You say in a commit message

> Using GIFs with `tasty-golden` for testing mostly works well, except that it has some issues displaying the `ByteString` in the terminal when the output differs from what is expected (outputs `Exception: Cannot decode byte`). But this is fine, since what we'd want to do in such a situation is to just pass `--accept` and compare the old and new GIF as rendered.

Two points here:
- you don't seem to have actually added any tests
- this seems like an odd problem, as `tasty-golden` treats files as binary. However, I think it is actually the upstream issue https://github.com/UnkindPartition/tasty-golden/issues/51. If you agree, it is perhaps worth linking that

primer/src/Primer/Typecheck.hs
```diff
@@ -744,7 +744,7 @@ check t = \case
             scrutWrap <- Hole <$> meta' (TCSynthed (TEmptyHole ())) <*> pure (addChkMetaT (TEmptyHole ()) e')
             pure $ Case caseMeta scrutWrap [] CaseExhaustive
       Left (TDIPrim tc) -> do
-        unless (tc == tInt || tc == tChar) $ throwError' $ InternalError $ "Unknown primitive type: " <> show tc
+        unless (tc == tInt || tc == tChar || tc == tAnimation) $ throwError' $ InternalError $ "Unknown primitive type: " <> show tc
         let f b = case caseBranchName b of
               PatCon _ -> Nothing
               PatPrim pc -> case pc of
```
brprice, line 750: It might be nice here to link to a note giving an overview of the `Animation` type. In particular to explain why we do not allow pattern matching on it

georgefst: I added this line yesterday to fix a test failure\*, without initially looking at the context, which I was going to ask you about. But anyway, I've worked it out now, and realised the logic was actually wrong, and have now fixed this in 76d51cc.

I've also rewritten the whole passage in a form which I consider much easier to follow, and which would extend more easily to adding further primitives.

\* This can be hit when `cabal run primer-test -- -p "tcWholeProg idempotent" --hedgehog-replay "375/638:oAbA2d Seed 3320906864796586236 8917501983448077275"`

georgefst:
> I believe you said at our last meet-up that a Primer primitive for `float` (or similar) might be helpful in this API. Can you comment briefly here in the discussion thread on where/how?

This is mentioned several times in #1173. Essentially, all mentions of integer in this PR should be replaced with a real number representation, freeing students from thinking in terms of discrete time steps and drawing on fixed grid lines.

georgefst:
Turns out I pushed the wrong branch after a rebase, which explains the missing tests, the empty module, and the negations on `rectEnvelope` being in the wrong commit. Sorry!

I've now applied all suggestions, and filled in #1169. I'm happy to merge this if there are no further objections.

primer/src/Primer/Typecheck.hs
brprice: Thanks for refactoring this, the new version is indeed much clearer

PR #1167, opened 2023-10-24
test: multiple requests succeed

brprice:
We run multiple requests (actions, evaluation, questions, undo, redo etc) and check they are successful, or at least fail in expected ways.

### DIFF
--- a/primer/src/Primer/App.hs
+++ b/primer/src/Primer/App.hs
@@ -678,6 +678,7 @@ applyProgAction prog = \case
     let def = ASTDef expr ty
     pure (insertDef mod name $ DefAST def, Just $ SelectionDef $ DefSelection (qualifyName modName name) Nothing)
   AddTypeDef tc td -> editModule (qualifiedModule tc) prog $ \m -> do
+    when (Map.member (baseName tc) (moduleTypes m)) $ throwError $ TypeDefAlreadyExists tc
     td' <- generateTypeDefIDs $ TypeDefAST td
     let tydefs' = moduleTypes m <> Map.singleton (baseName tc) td'
     liftError
--- a/primer/src/Primer/App.hs
+++ b/primer/src/Primer/App.hs
@@ -29,6 +29,8 @@ module Primer.App (
   runEditAppM,
   runQueryAppM,
   Prog (..),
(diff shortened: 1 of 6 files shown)


brprice:
This supercedes #973

primer/src/Primer/App.hs
```diff
@@ -1494,17 +1494,17 @@ checkProgWellFormed p =
...
 
--- | Construct a new, empty expression
+-- | Construct a new, empty expression (with typecache of TEmptyHole, both synthed and checked)
 newExpr :: MonadFresh ID m => m Expr
 newExpr = do
   id_ <- fresh
-  pure $ EmptyHole (Meta id_ Nothing Nothing)
+  pure $ EmptyHole (Meta id_ (Just $ TCEmb $ TCBoth (TEmptyHole ()) (TEmptyHole ())) Nothing)
 
--- | Construct a new, empty type
```
dhess, line 1496: It's not obvious (to me, anyway) what this `fix` commit fixes. I can see that a change has been made to `newType` and `newExpr`, and a corresponding test change, but since presumably the test was already passing before this change, it's all a bit circular.

Can you add a short comment to the commit about what, precisely, this change fixes?

primer/test/Tests/Action/Prog.hs
```diff
@@ -393,7 +393,7 @@ unit_create_typedef_bad_2 =
           , astTypeDefNameHints = []
           }
    in progActionTest defaultEmptyProg [AddTypeDef (tcn "T") td1, AddTypeDef (tcn "T") td2]
-        $ expectError (@?= TypeDefError "InternalError \"Duplicate-ly-named TypeDefs\"")
+        $ expectError (@?= TypeDefAlreadyExists (tcn "T"))
```
dhess, line 396: Good idea.

dhess (approved):
Thanks! Cleanup like this isn't very glamorous or fun, but it's much appreciated.

primer/src/Primer/App.hs
brprice: Thanks & done. This got me to notice the same problem exists for a new session, which I have also fixed in the same way -- see ee21d0c9

Issue #81, opened 2020-12-16
Improved inference


dhess:
(Note: I'm not sure the title is quite right. Feel free to change it.)

Consider the following partial program:

<img width="1276" alt="Screen Shot 2020-12-16 at 1 57 24 PM" src="https://user-images.githubusercontent.com/2438/102361084-bc3c7100-3faa-11eb-9382-733eea9c4a1f.png">

That all looks good. Now let's change it a bit:

<img width="1276" alt="Screen Shot 2020-12-16 at 1 54 26 PM" src="https://user-images.githubusercontent.com/2438/102361265-f3128700-3faa-11eb-83ae-1206f5b2f16a.png">

Here I've changed the input type of the lambda from `Bool` to `?`, and also changed the argument in the application from `?` to `False`. In the second program, Vonnegut knows that the hole type `?` can only be filled with `Bool` -- it displays an error if I try to fill it with `Nat` -- but it doesn't tell me that it knows this when I select that hole. It shows "Kind: Type", but what I want it to show is something along the lines of "Expected type: Bool, Actual type: ?", just as it did in the first program when the holes were swapped.

brprice:
Another facet of improved inference is to surface information about consistent-type constraints to the student. For example
- `foo : T; foo = ? True` checks and says the hole synthesises type `?`. This is correct, but it may be nice to know that to complete the program one needs to fill it with a term of type `Bool -> T`. (Reporting `? -> T` is probably an easier first step, since this more closely matches the current "top-down, left-to-right" typechecking algorithm)
- `foo : T; foo = ? @Bool` (a type application) checks and says the hole synthesises type `?`. Similar to the above this is fine, but would be nice to know it needs filling with a term of type `∀α. S[α]` for some `S` such that `S[Bool] = T`. (Note that it is impossible to determine `S[α]` from solely knowing `S[Bool] = T`. It is not clear what should be reported to a student here -- explicitly saying things about an unknown `S` may be confusing, maybe we should just say `∀_.?`, or simply "some ∀ type"?
- `foo : ? Bool; foo = foo` checks and says the (type) hole synthesises the kind `?`. It may be nice to know  `* -> *` is needed. (Here we know it must be an arrow since it occurs in function position, and its source is `*` as it is applied to `Bool : *`, and its target is `*` as `? Bool` is the type of a term and thus must be of kind `*`)

The above all have analogues for non-empty holes / mismatches.

dhess:
[...]

PR #1172, opened 2023-11-08
feat!: API gives more details on mismatched types

brprice:
For non-empty holes (aka "mismatches"), the OpenAPI now gives both the expected and actual types. We also record the checked-at kind of non-empty type holes in the metadata rather than the synthesised one, for consistency with the term level.

BREAKING CHANGE: this changes the OpenAPI schema.

### DIFF
--- a/primer-api/src/Primer/API.hs
+++ b/primer-api/src/Primer/API.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE OverloadedLabels #-}
 {-# LANGUAGE OverloadedRecordDot #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -33,6 +34,7 @@ module Primer.API (
   findSessions,
   getVersion,
   Tree,
+  treeIds,
   NodeBody (..),
   viewProg,
   Prog (Prog),
@@ -71,6 +73,7 @@ module Primer.API (
   undoAvailable,
   redoAvailable,
   Name (..),
+  OkOrMismatch (..),
   TypeOrKind (..),
   getSelectionTypeOrKind,
 ) where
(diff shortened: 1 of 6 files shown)


brprice:
This PR works pretty well for terms, but for types it is somewhat lacking: in `List {? Maybe ?}` we will report that the mismatch has actual type `* -> *` (good), but say we expect `?` (a kind hole). This makes sense given the kindchecker's behaviour, but we could probably do a better job with the metadata there -- I'll investigate.

We also have a more tricky(?) issue with `{? Nat ?} Bool`, where the mismatch has actual type `*` (good -- this is the kind of `Nat`), and expected type `?` (misleading -- we expected an arrow kind).

brprice:
> This PR works pretty well for terms, but for types it is somewhat lacking: in `List {? Maybe ?}` we will report that the mismatch has actual type `* -> *` (good), but say we expect `?` (a kind hole). This makes sense given the kindchecker's behaviour, but we could probably do a better job with the metadata there -- I'll investigate.
> 
> We also have a more tricky(?) issue with `{? Nat ?} Bool`, where the mismatch has actual type `*` (good -- this is the kind of `Nat`), and expected type `?` (misleading -- we expected an arrow kind).
> 
> Similarly with the term `{? Zero ?} True` we report we expected `?`, but we may prefer to say `? -> ?` (or even `Bool -> ?`, but that is heading to #81 territory)

I've pushed a draft commit that does this for kinds, but not types. The remaining cases to consider are "a term hole in SYN position where the typing rule has a `matchArrow/Forall` constraint" where we could say `? $ True` expects an arrow type, and `? @Bool` expects a forall type. Reporting `? -> ?` and `∀α:?.?` I think is fairly straightforward, but being smarter and constraining those holes is best left to #81.  (Note that this includes saying `? x y` expects type `? -> ? -> ?` rather than just `? -> ?`)

brprice:
On further thought, I have decided to not include any inference/constraint propagation in this PR. Mismatches in synthesisable position which nonetheless have constraints will simply say "expected a hole type", e.g. in `? $ True`, the hole in function position won't say that it requires an arrow type. Further work is left to #81. However we will ensure to treat non-empty type holes in "checkable" position similarly to term holes even though our type level is synthesis-only rather than properly bidirectional.

The reasoning here is 
- let's keep this pr simple so we can merge soon
- further work is lower priority than getting a basic version
- doing better is difficult to do consistently. Consider
  - `{? 1 ?} True`, we could say `expected: ? -> ?`, but why not `Bool -> ?`
  - `? True` should probably have the same "expected" as the non-empty hole version, but this PR is only focusing on displaying the mismatch for non-empty holes
  - `{? 1 ?} True ? Nil` should probably say `Bool -> ? -> List ?` (or maybe `? -> ? -> ?`), but actually reporting this is non-trivial without properly representing and propagating constraints (since this expression is curried, i.e. is actually two nested applications, and thus knowing we need two arrows in the type needs some slightly non-local information)

georgefst (approved):
Looks good.

This is obviously useful even without (some form of) #81, so I agree with merging in its current state, much as some outputs are unsatisfying.

PR #1174, opened 2023-11-15
chore: switch from `uuid` to `uuid-types`

dhess:
`uuid` drags in `network-info`, which doesn't build on GHC's Wasm target. The only drawback to `uuid-types` for our purposes is that it doesn't include a cryptographically secure generator for UUIDv4, which we've been using up until now (specifically, the `nextRandom` effect).

Therefore, we replace `nextRandom` with our own pseudo-random implementation. Note that our implementation is *not* cryptographically secure, as `entropy` does not build on the Wasm target, either.

### DIFF
--- a/primer/src/Primer/UUIDv4.hs
+++ b/primer/src/Primer/UUIDv4.hs
@@ -1,0 +1,14 @@
+module Primer.UUIDv4 (nextRandom) where
+
+import Foreword
+
+import Data.UUID.Types (UUID)
+import System.Random (randomIO)
+
+-- | Generate a pseudo-random 'UUID'.
+--
+-- Note: this effect does __not__ generate cryptographically secure
+-- UUIDs. If you need cryptographically secure 'UUID's, use
+-- the effect of the same name from the @uuid@ package.
+nextRandom :: IO UUID
+nextRandom = randomIO
--- a/primer-api/primer-api.cabal
+++ b/primer-api/primer-api.cabal
@@ -149,7 +149,7 @@ test-suite primer-api-test
       , tasty-hunit
       , text
       , transformers
-      , uuid
+      , uuid-types                                        ^>=1.0.5.1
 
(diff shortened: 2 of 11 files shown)


dhess:
Turns out that `entropy` doesn't build with the Wasm target, either, so I've reverted to an earlier implementation of `nextRandom` from @brprice. Note that this means our `nextRandom` is no longer cryptographically secure, but we're not relying on that, anyway.

primer/src/Primer/UUIDv4.hs
```diff
@@ -0,0 +1,18 @@
...
+-- | Generate a pseudo-random 'UUID'.
+--
+-- Note: this effect does __not__ generate cryptographically secure
+-- UUIDs. If you need cryptographically secure 'UUID's, use
+-- the effect of the same name from the @uuid@ package.
+nextRandom :: IO UUID
+nextRandom = do
+  [b0, b1, b2, b3, b4, b5, b6, b7, b8, b9, ba, bb, bc, bd, be, bf] <-
+    replicateM 16 randomIO
+  pure $ buildFromBytes 4 b0 b1 b2 b3 b4 b5 b6 b7 b8 b9 ba bb bc bd be bf
```
brprice, line 18: Another option would be something like `nextRandom = randomIO`, since `UUID` has a `Random` instance' which is essentially the same implementation, I think.

brprice (approved):
Minor comments above. It may be worth explicitly stating in the commit message(?) that we don't (or don't plan to, once have proper auth?) rely on the cryptographic security property anywhere.

primer/src/Primer/UUIDv4.hs
dhess: yeah, I think I'll switch to `randomIO`

dhess: Done.

PR #1176, opened 2023-11-22
feat: Add flavor annotations to edges in API trees

georgefst:
This will enable more detailed rendering by clients.

We have opted to keep `EdgeFlavor` small, in order to make things easier for clients. We _might_ eventually want to separate e.g. type- and term-level analogues in to separate flavors, but hopefully what we have here is enough information, since we anticipate these only being used for simple labelling.


primer-api/src/Primer/API/RecordPair.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE NoFieldSelectors #-}
+
```
brprice, line 2: Why is this pragma needed?

brprice (approved):

primer-api/src/Primer/API/RecordPair.hs
georgefst: We need to import the type fully in `Primer.API` to use the generic lenses. But `fst` and `snd` as selectors obviously cause a lot of name clashes.

Issue #1133, opened 2023-09-05
How to avoid deleting students' work


georgefst:
There are several scenarios where an action other than "delete this node" could cause the student to lose an entire subtree of unbounded size. We have at least:

- Deleting a typedef or value constructor. See #401.
  - ~~We currently disable this for in-use typedefs, since the simple way to fix up the program would then be to delete all subtrees mentioning that type or constructor, and we don't want to delete students' work, at least not without making it very clear to them what is happening.~~ EDIT: relaxed in #1153
- Deleting an in-use function. See https://github.com/hackworthltd/primer-app/issues/952.
  - ~~We disable this for similar reasons to the typedefs case above. It's easier to see how we might fix up the program here (replace variable uses with holes), but it can still have non-local effects, where it would not be immediately obvious what has changed in a large program.~~ EDIT: relaxed in #1182
- Once #402 is implemented, deleting a type parameter.
- Case branches. See `Scrutinee type changes in match _ with expressions` in #132.
  - This differs from the previous cases in that it's local. Which is probably why we do currently allow it.
  - Note that changing scrutinee type is no longer the _only_ way in which pattern branches can be removed, as now that we support wildcards we can also remove branches directly, but given that this involves an action which is labelled as being destructive, this isn't really relevant here since it's similar to the "delete this node" case.

## Possible solutions

We have discussed a range of ideas for handling this issue:

- Keeping nonsensical ASTs in the program and exempting them from typechecking: https://github.com/hackworthltd/primer/issues/118.
  - Or some sort of special bin for trees which once made sense but, for example, mention non-existent types.
- Always duplicate the old typedef/function/branch. This is inspired by Unison but may not work well without embracing the rest of Unison's design. This might have worked better if we had implemented our modules-as-nested-records idea.
- Just allowing all sorts of edits and merely warning the student somehow.

dhess:
As discussed in today's wrap-up meeting:

* Allow all edits and show a warning is a good starting place, and probably good enough for Primer 1.0
* Later versions of Primer might add constructs to keep around code that no longer typechecks, for reference.
* One potentially interesting point in the design space would be a time machine-like view where you could put the current version of any definition side-by-side with older versions, and go back in "edit time" to compare the current with previous versions. Then, either the student could crib off the older version to restore code that had been deleted by, say, a value constructor having been removed or changed; or possibly we could implement some kind of copy-paste action that would try to paste a subtree from a previous version of the definition into the current definition.

[...]

PR #1182, opened 2023-11-28
Allow deleting an in-use term definition

georgefst:
For consistency with https://github.com/hackworthltd/primer/pull/1153, as anticipated in https://github.com/hackworthltd/primer/issues/1133#issuecomment-1715843100.

The only context in which we still ban edits rather than fixing up use sites is for type definition parameters, as tracked in https://github.com/hackworthltd/primer/issues/402.

### DIFF
--- a/primer/src/Primer/Def/Utils.hs
+++ b/primer/src/Primer/Def/Utils.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE OverloadedLabels #-}
 
-module Primer.Def.Utils (nextID, nextIDTypeDef, globalInUse, typeInUse) where
+module Primer.Def.Utils (nextID, nextIDTypeDef, typeInUse) where
 
 import Foreword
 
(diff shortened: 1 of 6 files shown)


primer/src/Primer/Def/Utils.hs
```diff
@@ -40,12 +40,6 @@ nextIDTypeDef (TypeDefPrim (PrimTypeDef ps _)) =
   succ $ foldlOf' (folded % _2 % kindIDs) max minBound ps
 {-# INLINE nextIDTypeDef #-}
 
-globalInUse :: Foldable f => GVarName -> f Def -> Bool
-globalInUse v =
-  anyOf
-    (folded % #_DefAST % #astDefExpr % to freeGlobalVars)
-    (Set.member v)
-
```
georgefst, line 48: Alternatively, we could mark this as a Weeder root.

brprice: I don't think there is much point in keeping this around. It can always be fished out of the repo's history if needed.

brprice (approved):

PR #1186, opened 2023-11-29
chore: Disable animations support

georgefst:
@dhess


dhess:
I added the following note to the commit message, which I'll repeat here for better discoverability:

Specifically, we remove our dependencies on `diagrams-lib` and `diagrams-rasterific`, which are blocking support for Wasm targets. As these dependencies were only workarounds until we can design a pure animation system, we're not losing *too* much here.

PR #1187, opened 2023-11-29
feat: interpreter (faster evaluator-to-normal-form)

brprice:
The one feature of the old stepwise evaluator that we don't replicate is the ability not to evaluate under lambdas.
This may be annoying if this new interpreter is hooked up to a frontend, as then any evaluation of a recursive function will diverge, even if an evaluation of that same function applied to an argument may succeed. The possible mitigations for this include
- only use this new interpreter for things of "base" type, i.e. things which have a type such that we expect them not to contain un-applied lambdas (note that if they contain holes, this expectation may be false, so a refinement would be "base type and not contain holes"). This is not quite "not an arrow/forall type", as we need to detect things like `Maybe (Bool -> Char)`, or a return type of `T` where `data T = C (Bool -> Char)`.
- revisit this new interpreter to avoid the hard requirement that it evaluates under lambdas. This can be done by stepping outside of the AST domain and representing the interpretation of a lambda as a closure (a pair of the current environment and the body), and then applications will require the function-position to be a closure rather than a syntactic lambda. (This may also be useful if we want to make `Expr` spine-strict, since we need to step outside of the AST domain to keep laziness in that case anyway.)

When working on this PR, at one point we had an explicit
"fuel" counter so that the interpreter would not loop on a divergent term.
This has been replaced (and rebased) with an IO-based System.Timeout call,
but the old state is still in the history for https://github.com/hackworthltd/primer/tree/brprice/wip/interp.

The reason we changed track here is
- We need to lazily produce the resultant AST, since the semantics of
  Primer (the object language) are lazy
  (in particular, code like `head (letrec l = Cons True l in l)` should
  reduce to `True` (ignoring type annotations)
- This means we can't use `ExceptT` (i.e. `Either`) to signal the
  "recursion depth exceeded" error, since to pattern match on the (root
  of the) AST of a recursive call would force the `Either`, and thus
  divergent subterms would throw a `Left`, aborting the whole computation.
- We can however throw an "imprecise exception" (e.g. Prelude's `error`,
  or `Control.Exception`s `throw`), from pure code. This is essentially
  hiding an exception deep within a (recursively returned) AST, and will
  not actually trigger unless it is forced.
- However, to make this testable, we need to be able to catch the
  "timeout" error in the testsuite, meaning we need to be able to run IO
  actions in Hedgehog tests, and thus need to refactor our WT monad to
  (optionally) include IO at the base.
- Having done this, it is just as easy to not have the recursion-depth
  check and use `System.Timeout` instead, which also works in `IO`,
  and avoids some overhead due to not tracking the depth (and makes the
  interpretation of recursive let bindings nicer.

Note that (c.f. comment in code) this new interpreter relies heavily on laziness and will break if `Expr` is made spine-strict.

### DIFF
--- a/primer/src/Primer/EvalFullInterp.hs
+++ b/primer/src/Primer/EvalFullInterp.hs
@@ -1,0 +1,437 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Primer.EvalFullInterp (
+  interp,
+  Timeout (MicroSec),
+  interp',
+  InterpError (..),
+  Dir (..),
+  mkEnv,
+  mkGlobalEnv,
+) where
+
+import Foreword
+
+import Control.Exception (throw)
+import Data.Map.Lazy qualified as Map
+import Data.Set qualified as Set
+import Primer.Core (
+  Bind' (Bind),
+  CaseBranch' (CaseBranch),
+  CaseFallback' (CaseFallback),
+  Expr' (..),
+  GVarName,
+  LVarName,
+  LocalName,
+  Pattern (PatCon, PatPrim),
+  PrimCon,
+  TmVarRef (GlobalVarRef, LocalVarRef),
+  TyConName,
+  TyVarName,
+  Type' (..),
+  ValConName,
+  bindName,
+  caseBranchName,
+  mapFallback,
+  unLocalName,
+ )
+import Primer.Core.DSL.Meta (create')
+import Primer.Core.Transform (decomposeTAppCon, unfoldApp)
+import Primer.Core.Utils (
+  concreteTy,
+  forgetMetadata,
+  forgetTypeMetadata,
+  freeVars,
+  freeVarsTy,
+  freshen,
+ )
+import Primer.Def (ASTDef (ASTDef), Def (DefAST, DefPrim), DefMap)
+import Primer.Eval.Redex (
+  Dir (Chk, Syn),
+ )
+import Primer.Name (Name)
+import Primer.Primitives (primConName, primFunDef)
+import Primer.Primitives.PrimDef (PrimDef)
+import Primer.TypeDef (
+  ASTTypeDef (ASTTypeDef),
+  TypeDef (TypeDefAST),
+  TypeDefMap,
+  ValCon (valConArgs, valConName),
+ )
+import System.Timeout (timeout)
+
+-- Invariant: vars is the set of free vars in the values of the map
+-- Thus the binders we need to rename before going under them.
+-- Invariant: the values in the env will be in normal form.
+data EnvTm = EnvTm
+  { vars :: Set Name
+  , env :: Map.Map (Either GVarName LVarName) (Expr' () () ())
+  , prims :: Map GVarName PrimDef
+  }
+data EnvTy = EnvTy
+  { vars :: Set Name
+  , env :: Map.Map TyVarName (Type' () ())
+  }
+
+-- | Convert an environment into the form needed for 'interp'
+mkEnv ::
+  [(Either GVarName LVarName, Expr' a b c)] ->
+  Map GVarName PrimDef ->
+  [(TyVarName, Type' a b)] ->
+  (EnvTm, EnvTy)
+mkEnv tms prims tys =
+  extendTmsEnv
+    (second forgetMetadata <$> tms)
+    ( EnvTm mempty mempty prims
+    , extendTysEnv' (second forgetTypeMetadata <$> tys) $ EnvTy mempty mempty
+    )
+
+-- | Convert an environment into the form needed for 'interp'
+mkGlobalEnv :: DefMap -> (EnvTm, EnvTy)
+mkGlobalEnv defs =
+  mkEnv
+    ( mapMaybe
+        ( \(f, d) -> case d of
+            DefAST (ASTDef tm ty) -> Just (Left f, Ann () (forgetMetadata tm) (forgetTypeMetadata ty))
+            _ -> Nothing
+        )
+        $ Map.assocs defs
+    )
+    ( Map.mapMaybe
+        ( \case
+            DefPrim p -> Just p
+            _ -> Nothing
+        )
+        defs
+    )
+    mempty
+
+data InterpError
+  = Timeout
+  | NoBranch (Either ValConName PrimCon) [Pattern]
+  | UnknownTyCon TyConName
+  | UnknownValCon TyConName ValConName
+  deriving stock (Eq, Show)
+  deriving anyclass (Exception)
+
+newtype Timeout = MicroSec Int
+
+-- | Wrap the interpreter in a IO-based timeout, and catch 'InterpError' exceptions
+interp ::
+  Timeout ->
+  TypeDefMap ->
+  (EnvTm, EnvTy) ->
+  Dir ->
+  Expr' () () () ->
+  IO (Either InterpError (Expr' () () ()))
+interp (MicroSec t) tydefs env dir e = do
+  e' <- timeout t (try $ evaluate $ force $ interp' tydefs env dir e)
+  pure $ case e' of
+    Nothing -> Left Timeout
+    Just e'' -> e''
+
+{- HLINT ignore interp' "Avoid restricted function" -}
+-- (we are intentionally using 'throw' here
+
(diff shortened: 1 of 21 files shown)


brprice:
@dhess: I'll leave some cleanup work for you to look at:
- testsuite should be DRY'd between Interpreter and FullStep (currently I've copy-pasted the file and hacked the tests, removing some that are not super relevant).

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,437 @@
...
+    --  @let xs=as:(lettype p=A in S) in e@ for @data T p = C S@
+    case interp' tydefs env Syn e of
+      Ann _ (Con _ c as) (decomposeTAppCon -> Just (tycon, tyargs))
+        | Just (CaseBranch _ xs t) <- find ((PatCon c ==) . caseBranchName) brs ->
+            let envTy' = extendTysEnv' (tyParamEnvExt tycon tyargs) envTy
+             in interp'
+                  tydefs
+                  ( extendTmsEnv
+                      ( zip (Right . bindName <$> xs)
+                          $ zipWith (\a argTy -> Ann () a $ interpTy envTy' argTy) as
```
brprice, line 268: TODO: I should probably evaluate the `as`?

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,437 @@
...
+      Ann _ (PrimCon _ c) (TCon _ ((== primConName c) -> True))
+        | Just (CaseBranch _ [] t) <- find ((PatPrim c ==) . caseBranchName) brs -> interp' tydefs env Chk t
+        | CaseFallback t <- fb -> interp' tydefs env Chk t
+        | otherwise -> throw $ NoBranch (Right c) $ caseBranchName <$> brs
+      -- literals (primitive constructors) are actually synthesisable, so may come
+      -- without annotations
+      PrimCon _ c
+        | Just (CaseBranch _ [] t) <- find ((PatPrim c ==) . caseBranchName) brs -> interp' tydefs env Chk t
+        | CaseFallback t <- fb -> interp' tydefs env Chk t
+        | otherwise -> throw $ NoBranch (Right c) $ caseBranchName <$> brs
```
brprice, line 286: This "duplication" is a bit of a worry. It is reflecting the same in the stepwise evaluator, but we should perhaps review the calculus to see if it is really needed.

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,444 @@
...
+-- subterms are local recursive helper functions (since we evaluate under
+-- lambdas and inside let bindings).
+--
+-- We have made the design choice to use imprecise exceptions to be able to
+-- write simple code which can report obvious problems without compromising
+-- laziness. Another choice would be to not report errors (and letting these
+-- subterms be stuck). A non-option would be to use ExceptT to report errors,
+-- since that would destroy our critical laziness properties: the monadic
+-- bind would force the top Left/Right constructor, and that in turn would
+-- force the full evaluation of the relevent subterms. This design choice
```
georgefst, line 179:
```suggestion
-- force the full evaluation of the relevant subterms. This design choice
```

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,444 @@
...
+-- incurring unbounded work. Note that this means we cannot make our 'Expr'
+-- type spine-strict without refactoring this interpreter!
+
+{-
+Future work:
+- can we reuse Redex module
+- perhaps do full NBE
+- maybe don't stay in AST domain
+  this perhaps may help with a pure recursion-depth limit and reporting partial results?
+-}
```
georgefst, line 199: We should expand on this, maybe in a GitHub issue.

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,444 @@
...
+      _ -> throw $ UnknownTyCon tcon
+    ann d e t = upsilon d $ Ann () e t
+    upsilon d = \case
+      Ann _ e t ->
+        -- We force t before emitting anything, so this is not very lazy.
+        -- However we know evaluation of types must terminate!
+        case (d, concreteTy t) of
+          (Chk, True) -> upsilon d e
+          _ -> Ann () e t
+      e -> e
```
georgefst, line 321: This function could do with an explanatory comment.

primer/src/Primer/EvalFullInterp.hs
```diff
@@ -0,0 +1,444 @@
...
+freshLike v (envTm, envTy) =
+  let avoid =
+        envTm.vars
+          <> envTy.vars
+          <> Set.delete
+            (unLocalName v)
+            ( Set.fromList (fmap unLocalName $ rights $ Map.keys envTm.env)
+                <> Set.map unLocalName (Map.keysSet envTy.env)
+            )
+   in freshen avoid v
```
georgefst, line 334: This function could do with an explanatory comment.

primer/src/Primer/EvalFullInterp.hs
brprice: See #1194

brprice:
I think this is more-or-less ready, except the testsuite run seems to fail on linux CI (OOM??). I'll need to investigate further.

dhess:
(I think the last OOM was due to a long-running earlier test that was chewing up lots of resident RAM.)

brprice:
> I think this is more-or-less ready, except the testsuite run seems to fail on linux CI (OOM??). I'll need to investigate further.
> 
> Some initial investigation shows that running the CI testsuite (`nix run -v -L -j2 .#'primer:test:primer-test' -- -j1`) artifact with -j1 is fairly static at 700M resident memory, but this branch seems to quickly grow to multiple tens of gigabytes (even when running old tests unrelated to the interpreter). I am confused about this and will look again in the near future.

This OOM is due to 14096bf1 refactor: generalise WT to be a transformer.

(A useful technique here is to run `primer-test` with ghc rts options `+RTS -M1G` to limit the heap size. This PR will abort with an exhausted heap, and main will not. `git bisect run` can then automate testing each commit)

dhess:
`WT` is only used for testing, correct?

brprice:
> `WT` is only used for testing, correct?

Yes. It is used in property tests only

brprice:
> > I think this is more-or-less ready, except the testsuite run seems to fail on linux CI (OOM??). I'll need to investigate further.
> > Some initial investigation shows that running the CI testsuite (`nix run -v -L -j2 .#'primer:test:primer-test' -- -j1`) artifact with -j1 is fairly static at 700M resident memory, but this branch seems to quickly grow to multiple tens of gigabytes (even when running old tests unrelated to the interpreter). I am confused about this and will look again in the near future.
> 
> This OOM is due to [14096bf](https://github.com/hackworthltd/primer/commit/14096bf1beeebf998a00901a7ee1af797434acf5) refactor: generalise WT to be a transformer.
> 
> (A useful technique here is to run `primer-test` with ghc rts options `+RTS -M1G` to limit the heap size. This PR will abort with an exhausted heap, and main will not. `git bisect run` can then automate testing each commit)

I have force-pushed a fix for this. Essentially we don't make `WT` a transformer, but just change the base monad to IO. This does still need `TestM` to be transformer-ified. The problem seems to be GHC emitting less-efficient code for `WT` being a transformer, even though we only actually use it at `IO` -- this could probably be fixed with the correct pragmas, but that seems somewhat pointless.

I think this PR is ready to merge now.

@dhess: note that there are two `TODO`s left in the code that I will leave for you in future work. They are
- DRY the testsuites for `interp` and the old stepping full eval (as much as makes sense, at least)
- enable two more tests for `interp` on a prog with imported modules (needs `interp` hooked up to the core api layer)

dhess:
Thanks! The speedup here will make a huge impact.

primer/src/Primer/EvalFullInterp.hs
dhess: See https://github.com/hackworthltd/primer/issues/1237

dhess:
> @dhess: note that there are two `TODO`s left in the code that I will leave for you in future work. They are
> 
> * DRY the testsuites for `interp` and the old stepping full eval (as much as makes sense, at least)
> * enable two more tests for `interp` on a prog with imported modules (needs `interp` hooked up to the core api layer)

Note that the first of these was (at least partially) addressed in #1238, and the second is addressed in #1241.

PR #1190, opened 2023-11-30
feat: build & run `primer-benchmark` for Wasm32 targets

dhess:
Note: there's still quite a bit of work to do on this, because `criterion` and at least one of its dependencies don't work with the Wasm backend target. I'll take a look at [tasty-bench](https://github.com/Bodigrim/tasty-bench), which explicitly states that it supports Wasm in its README.


dhess:
Closing, will be fixed by #1429, and we didn't need to port anything as Criterion works in Wasm now.

PR #1191, opened 2023-11-30
fix: make Wasm upstream overrides dependent on `arch(wasm32)`

dhess:
This fixes an issue where Nix caching was significantly degraded in the default Nix shell, due to haskell.nix not being able to build `source-repository-package`s using Nix in Nix shells. We'd prefer not to do this, because it means that our Wasm dependencies are ever so slightly different than our native target dependencies, but at the moment the differences should be small enough not to matter.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -45,32 +45,40 @@ source-repository-package
   --sha256: 0q8m8asmb83mpa3j3adlrhc446xif7gv6lql20gv05k33lmbjfhg
 
 -- Wasm workarounds.
+--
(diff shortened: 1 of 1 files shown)


brprice:
This fixes the symptom I was seeing locally: `nix develop -v -L -j2  .?ref=afbb312 -c bash -c 'ghc-pkg list | grep aeson'` now shows `aeson` is in our devshell's cabal store

dhess:
This creates a regression in the Wasm build, but in the interest of fixing everyone's Nix shell, I'm going to commit it anyway and fix the Wasm regression later.

PR #1226, opened 2024-04-08
ci: only run Wasm tests once a PR reaches the merge queue

dhess:
The Wasm tests are very expensive, and our build system isn't smart enough to know when they don't need to be re-run, so for now, we only run the tests once a PR hits the merge queue.

### DIFF
--- a/.buildkite/primer-wasm.yaml
+++ b/.buildkite/primer-wasm.yaml
@@ -1,10 +1,30 @@
+# Note: because they take so much time to run, we only run the Primer
+# Wasm tests when one of these conditions is true:
+#
+# 1. The PR has hit the merge queue.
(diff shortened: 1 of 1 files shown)


dhess:
Note that this implementation will still *build* the Primer Wasm test suite in all cases; it just won't run them except under the conditions explained in the commit message. I think this will be good enough.

dhess:
Hmm, adding the label didn't cause Buildkite to run the tests &mdash; it took the `build.branch !~ /^gh-readonly-queue\// && !(build.pull_request.labels includes "Run Wasm tests")` conditional.

I'll need to look into this, but this isn't sufficient to block the merge. First and foremost, we need not to run the Wasm tests on every PR/push.

PR #1233, opened 2024-04-11
chore(hackage): bump dependencies to latest versions

dhess:


dhess:
Hmm, these relatively minor version bumps are causing the Wasm build to fail, apparently because it's trying to build `unix`, which we don't directly depend on, and which we tell `tasty` to disable. But as we've only bumped `text`, `time`, and `containers`, none of which depend on `unix` either, I'm not sure what has happened.

dhess:
Ugh, this was mistakenly merged to main :\ I think I'm going to rewrite history. Sorry :(

PR #1234, opened 2024-04-11
chore(hackage): upstream `semirings` supports Wasm now

dhess:
However, it hasn't been updated on Hackage yet, so we still need to resort to a `source-repository-package` for now.


dhess:
We should be able to retire our `semirings` pseudo-fork now.

Issue #1239, opened 2024-04-15
Hook interpreter up to API


dhess:
Will be closed by #1241.

PR #1241, opened 2024-04-15
feat: expose the interpreter via the various APIs

dhess:
Note that we only expose the time-bounded interpreter via the HTTP API, because the non-bounded variant would enable trivial DoS attacks against a Primer service. We do, however, expose the non-bounded variant via the local API, as it might be useful to handle the various exceptions it can throw (and runaway evaluations) in an application-specific way, differently to how the time-bounded variant does it.

Closes https://github.com/hackworthltd/primer/issues/1239.

Note that there's currently a bug in the interpreter that prevents it from reducing top-level definitions that contain applications. See https://github.com/hackworthltd/primer/issues/1247.

### DIFF
--- a/primer-api/primer-api.cabal
+++ b/primer-api/primer-api.cabal
@@ -156,6 +156,7 @@ test-suite primer-api-test
     , stm-containers
     , tasty                   ^>=1.5
     , tasty-discover
+    , tasty-expected-failure  ^>=0.12.3
     , tasty-golden            ^>=2.3.5
     , tasty-hunit
     , text
--- a/primer-api/src/Primer/API.hs
+++ b/primer-api/src/Primer/API.hs
@@ -54,6 +54,12 @@ module Primer.API (
   evalFull,
   EvalFullResp (..),
   evalFull',
+  evalBoundedInterp,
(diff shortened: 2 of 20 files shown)


dhess:
Hmm, I've discovered during testing that the `primer-api` methods `evalInterp`, `evalInterp'`, `evalBoundedInterp`, and `evalBoundedInterp`' don't reduce top-level definitions containing applications. I'm adding some unit tests to demonstrate this now.

dhess:
Actually, it's not just the `primer-api` methods: `interp` doesn't reduce top-level definitions containing applications, either.

dhess:
The bug is now tracked in https://github.com/hackworthltd/primer/issues/1247

dhess:
I think I'm going to merge this after wrapping "expected" around the test failures. There's a substantial amount of work here, in addition to a non-trivial frontend PR in https://github.com/hackworthltd/primer-app/pull/1169, and since it may take some time for me to get up to speed on the interpreter implementation in order to find the issue, I don't want these PRs to languish. Though the functionality here is very nice to have (the interpreter is an order of magnitude or more faster than the step evaluator), we can live without it until I can get to the root of the problem.

PR #1264, opened 2024-06-02
chore: WIP GHC 9.10.1 support

dhess:
Most of our ancillary tools aren't yet GHC 9.10.1-compatible and need to be disabled to get this working.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -30,17 +30,7 @@ haddock-quickjump:        True
 haddock-hyperlink-source: True
 haddock-internal:         True
 
-allow-newer:
-  , hedgehog-classes:hedgehog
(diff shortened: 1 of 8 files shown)


dhess:
Hmm, this currently breaks on Protolude, of all things. It might be time to switch to Relude.

dhess:
I've done some preliminary work on porting to Relude, but it's going to require a significant number of changes. I think for now we'll use the Protolude 9.10 fork, and possibly even vendor it, since Protolude looks to be more or less unmaintained now.

If we do decide to switch to Relude, we should probably do it bit-by-bit rather than all at once, given the number of changes required.

dhess:
Closing in favor of #1274

PR #1274, opened 2024-07-18
chore: upgrade to GHC 9.10.1

dhess:
Note that Protolude upstream doesn't support GHC 9.10, so we use a (hopefully temporary) fork.


dhess:
Workarounds needed:

* [ ] Disable hlint: https://github.com/ndmitchell/hlint/pull/1594

PR #1304, opened 2024-11-24
feat: add `primer-miso`, an attempt to re-implement Primer's UI in Haskell

dhess:
This PR is a proof-of-concept re-implementation of our TypeScript- & React-based Primer frontend application in Haskell, using [Miso](https://haskell-miso.org) and GHC's Wasm backend. The goal of this PR is the following:

* Render an arbitrary Primer term definition (we've chosen polymorphic `map` here) as a tree, completely in the browser, with no backend service required.
* Upon selecting an arbitrary subexpression of the definition, render the subexpression's type as a tree alongside the definition's tree.
* Use as little JavaScript/TypeScript as possible.
* Make a best-effort attempt to mimic the CSS styling of the TypeScript/React implementation. (Notably, we have not yet implemented the "tidy trees" algorithm, nor have we made any attempt to render pattern matches left-to-right as the TypeScript/React implementation does.)

The experience has been very positive, so we plan to follow this work up with one or more additional PRs that will add basic editing, support for multiple definitions, and program evaluation. Assuming that additional work is as successful as this initial PR has been, we will likely deprecate the TypeScript/React frontend and replace it with a pure Haskell version.

Note that this PR is based on @georgefst's original work in https://github.com/georgefst/ghc-wasm-miso-examples/tree/primer.

Also note that the code contains a number of `TODO` comments. Normally we don't permit such comments without a link to a relevant GitHub issue, so that we don't lose track of work that remains to be done. However, in the case of this particular PR, we'll relax this rule, as it's not yet clear exactly how to resolve these `TODO`s, or even whether they'll remain relevant as we implement the full editing API into `primer-miso` in subsequent work.

### DIFF
--- a/primer-miso/Makefile.wasm32
+++ b/primer-miso/Makefile.wasm32
@@ -1,0 +1,26 @@
+# NOTE:
+#
+# This Makefile assumes you're using the `nix develop .#wasm` shell.
+
+CABAL = wasm32-wasi-cabal
+
+build:	configure
+	$(CABAL) build
+
+frontend: build
+	./build-frontend.sh
+
+frontend-prod: build
+	./build-frontend.sh -Oz
+
+serve-frontend: frontend
+	simple-http-server -i -p 8000 -- dist
+
+configure:
+	$(CABAL) configure
+
+clean:
+	$(CABAL) clean
+	rm -rf dist/
+
+.PHONY: build configure clean
--- a/primer-miso/Makefile.wasm32
+++ b/primer-miso/Makefile.wasm32
@@ -4,7 +4,7 @@
 
 CABAL = wasm32-wasi-cabal
 
-build:	configure
+build:
 	$(CABAL) build
 
 frontend: build
@@ -16,11 +16,8 @@ frontend-prod: build
 serve-frontend: frontend
 	simple-http-server -i -p 8000 -- dist
 
-configure:
-	$(CABAL) configure
-
 clean:
 	$(CABAL) clean
 	rm -rf dist/
 
-.PHONY: build configure clean
+.PHONY: build clean
--- a/primer-miso/build-frontend.sh
+++ b/primer-miso/build-frontend.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ $# -eq 0 ]; then
+    echo "Building for dev"
+    dev_mode=true
+else
+    echo "Building for prod"
+    dev_mode=false
+fi
+
+rm -rf dist
+mkdir dist
+cp frontend/*.html dist/
+
+hs_wasm_path=$(wasm32-wasi-cabal list-bin -v0 exe:primer-miso)
+
+ghc_wasm_jsffi="dist/ghc_wasm_jsffi.js"
+
+"$(wasm32-wasi-ghc --print-libdir)"/post-link.mjs \
+                                   --input "$hs_wasm_path" --output "$ghc_wasm_jsffi"
+
+if ! [ -f "$ghc_wasm_jsffi" ] ; then
+    echo "post-link.mjs didn't produce a $ghc_wasm_jsffi file. Make sure you're in the Nix Wasm shell."
+    exit 1
+fi
+
+if $dev_mode; then
+    cp "$hs_wasm_path" dist/bin.wasm
+else
+    wizer --allow-wasi --wasm-bulk-memory true --init-func _initialize -o dist/bin.wasm "$hs_wasm_path"
+    wasm-opt ${1+"$@"} dist/bin.wasm -o dist/bin.wasm
+    wasm-tools strip -o dist/bin.wasm dist/bin.wasm
+    brotli --rm --best dist/bin.wasm -o dist/bin.wasm.br
+    mv dist/bin.wasm.br dist/bin.wasm
+fi
+
+cp frontend/*.js dist
--- a/primer-miso/exe/Main.hs
+++ b/primer-miso/exe/Main.hs
@@ -1,0 +1,32 @@
+{-# LANGUAGE CPP #-}
+
+#ifdef wasi_HOST_OS
+
+module MyMain (main) where
+
+import Foreword
+
+--import GHC.Wasm.Prim
+import Language.Javascript.JSaddle.Wasm qualified as JSaddle.Wasm
+import Primer.Miso (start)
+
+foreign export javascript "hs_start" main :: IO ()
+
+main :: IO ()
+main = JSaddle.Wasm.run start
+
+#else
+
+module Main (main) where
+
+import Foreword
+
+import Language.Javascript.JSaddle.Warp
+import Primer.Miso (start)
+
+-- Note that `debug` works with `cabal repl` but not `cabal run`.
+-- The best workflow is to run `ghcid -c "cabal repl primer-miso" -W -T ':main'`.
+main :: IO ()
+main = debug 8000 start
+
+#endif
--- a/primer-miso/exe/Main.hs
+++ b/primer-miso/exe/Main.hs
@@ -6,7 +6,6 @@ module MyMain (main) where
 
 import Foreword
 
---import GHC.Wasm.Prim
 import Language.Javascript.JSaddle.Wasm qualified as JSaddle.Wasm
 import Primer.Miso (start)
 
--- a/primer-miso/exe/Main.hs
+++ b/primer-miso/exe/Main.hs
@@ -20,7 +20,7 @@ module Main (main) where
 
 import Foreword
 
-import Language.Javascript.JSaddle.Warp
+import Language.Javascript.JSaddle.Warp ( debug )
 import Primer.Miso (start)
 
 -- Note that `debug` works with `cabal repl` but not `cabal run`.
--- a/primer-miso/primer-miso.cabal
+++ b/primer-miso/primer-miso.cabal
@@ -1,16 +1,21 @@
-cabal-version: 3.0
-name:          primer-miso
-version:       0.8.0.0
-license:       AGPL-3.0-or-later
-license-file:  COPYING
-copyright:     (c) 2024 Hackworth Ltd
-maintainer:    src@hackworthltd.com
-author:        Hackworth Ltd <src@hackworthltd.com>
-stability:     experimental
+cabal-version:      3.0
+name:               primer-miso
+version:            0.8.0.0
+license:            AGPL-3.0-or-later
+license-file:       COPYING
+copyright:          (c) 2024 Hackworth Ltd
+maintainer:         src@hackworthltd.com
+author:             Hackworth Ltd <src@hackworthltd.com>
+stability:          experimental
 synopsis:
   A web frontend for building Primer programs, using the Miso framework
 
-category:      UI
+extra-source-files:
+  build-frontend.sh
+  frontend/**/*.html
+  frontend/**/*.js
+
+category:           UI
 
 library
   exposed-modules:
@@ -100,8 +105,3 @@ executable primer-miso
       , websockets
 
     ghc-options:   -threaded -rtsopts -with-rtsopts=-N
-
-  extra-source-files:
-    build-frontend.sh
-    frontend/**/*.html
-    frontend/**/*.js
--- a/primer-miso/primer-miso.cabal
+++ b/primer-miso/primer-miso.cabal
@@ -1,0 +1,107 @@
+cabal-version: 3.0
+name:          primer-miso
+version:       0.8.0.0
+license:       AGPL-3.0-or-later
+license-file:  COPYING
+copyright:     (c) 2024 Hackworth Ltd
+maintainer:    src@hackworthltd.com
+author:        Hackworth Ltd <src@hackworthltd.com>
+stability:     experimental
+synopsis:
+  A web frontend for building Primer programs, using the Miso framework
+
+category:      UI
+
+library
+  exposed-modules:
+    Primer.Miso
+    Primer.Miso.Colors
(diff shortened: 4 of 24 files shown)


georgefst:
This mostly looks great! Thanks for handling all the boring stuff around licensing, packaging etc.

> Merge the relevant code in Primer.Miso.Util into core Primer.

I'd say this can wait. It won't be completely trivial in all cases to work out where it should go and how it should fit within Primer's existing API.

> Reformat the code using our Fourmolu settings.

Moving to two-space indentation (the only change Primer makes from Fourmolu defaults) is of course necessary. The other changes are regressions really, caused by Fourmolu not knowing operator fixities. I've rebased on #1305 instead, which fixes this by adapting some config I had in my original `ghc-wasm-miso-examples`-based version.

> Fix hlint warnings.
> Eliminate some variable shadowing warnings.

Ah, yes I haven't seen these because, respectively, HLS HLint support is currently lagging, and `ghc-wasm-miso-examples` disables shadowing warnings (which, it turns out, I don't hate, but let's not open that can of worms). Anyway, I've pushed fixes for both.

primer-miso/Makefile.wasm32
```diff
@@ -0,0 +1,26 @@
...
+	$(CABAL) build
+
+frontend: build
+	./build-frontend.sh
+
+frontend-prod: build
+	./build-frontend.sh -Oz
+
+serve-frontend: frontend
+	simple-http-server -i -p 8000 -- dist
```
georgefst, line 17: Maybe we should add some instructions to the README for this at least? It wasn't immediately obvious that the command to run the production server is `nix develop .#wasm -c make -f Makefile.wasm32 serve-frontend`.

georgefst (changes requested):

primer-miso/Makefile.wasm32
georgefst: Alternatively we could presumably create an actual Nix output for this? Note that in the absence of Wasm support from Haskell.nix, we just get a `nix run .#primer-miso:exe:primer-miso` that doesn't actually do anything. This is because when we compile natively, we currently use a `main` function which only really works in GHCI...

georgefst: We could alter our use of Jsaddle to give us a usable native binary, but honestly I'm not sure it's useful.

What might be useful is moving the comment from `main` in to an actual make target for ease of development. And potentially substituting GHCId with [ghciwatch](https://github.com/mercurytechnologies/ghciwatch), which is meant to be a modern replacement.

georgefst: > What might be useful is moving the comment from `main` in to an actual make target for ease of development.

_Also_, the command needs modifying at least slightly due to the project structure. We now require `ghcid -c "cabal repl exe:primer-miso" -W -T ':main'`. This still isn't very useful though, since it only picks up changes to `Main.hs`...

What we really want is `ghcid -c "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" -W -T 'Main.main'`, or `ghcid -c "cabal repl --enable-multi-repl exe:primer-miso all" -W -T 'Main.main'`. This should work according to [here](https://www.well-typed.co.uk/blog/2023/03/cabal-multi-unit/) and [here](https://github.com/ndmitchell/ghcid/issues/208), and even does for [my minimal example project](https://github.com/georgefst/hello-hs/compare/master...multi-component). But for whatever reason, I find that file changes are never picked up, i.e. when I save a file GHCId doesn't actually do anything.

@dhess I'd be interested to see whether you can reproduce this. It's a pretty crucial part of the developer experience.

(FWIW, Ghciwatch doesn't support this yet, but it's on the radar: https://github.com/MercuryTechnologies/ghciwatch/issues/316.)

primer-miso/primer-miso.cabal
```diff
@@ -0,0 +1,107 @@
...
+      , jsaddle-warp
+      , warp
+      , websockets
+
+    ghc-options:   -threaded -rtsopts -with-rtsopts=-N
+
+  extra-source-files:
+    build-frontend.sh
+    frontend/**/*.html
+    frontend/**/*.js
```
georgefst, line 107: `Warning: primer-miso.cabal:104:3: Unknown field: "extra-source-files"`

I don't think this does anything in this position.

primer-miso/build-frontend.sh
```diff
@@ -0,0 +1,39 @@
...
+fi
+
+if $dev_mode; then
+    cp "$hs_wasm_path" dist/bin.wasm
+else
+    wizer --allow-wasi --wasm-bulk-memory true --init-func _initialize -o dist/bin.wasm "$hs_wasm_path"
+    wasm-opt ${1+"$@"} dist/bin.wasm -o dist/bin.wasm
+    wasm-tools strip -o dist/bin.wasm dist/bin.wasm
+    brotli --rm --best dist/bin.wasm -o dist/bin.wasm.br
+    mv dist/bin.wasm.br dist/bin.wasm
```
georgefst, line 36: This doesn't work. Try changing `serve-frontend: frontend` to `serve-frontend: frontend-prod` in the makefile and you'll see the frontend fail to load, with something like `Uncaught CompileError: wasm validation error: at offset 4: failed to match magic number` in the browser console.

I think you need to set the right content headers somewhere. The ones which most servers enable by default for `gzip`.

georgefst: AFAICT, `simple-http-server` doesn't support this. But `http-server` from `npm` does: `nix shell nixpkgs/24.05#http-server -c http-server -b -p 8000 dist`.

georgefst: Also, I think one has to keep the `.br` suffix, or at least it's best practice (leaving it off does work in some cases for `gzip` at least).

dhess: (FYI, none of these problems are present on macOS, where the `prod` build works fine and serves up `.br` files without the suffix.)

dhess:
Bummer, I have my own fixes for most of this stuff that I worked on during a long flight, but hadn't yet committed to GitHub. Oh well. As all the checks pass, it looks like you've implemented most of the same fixes, so I'll review what you've done and add anything that is left over in my branch.

dhess:
> > Merge the relevant code in Primer.Miso.Util into core Primer.
> 
> I'd say this can wait. It won't be completely trivial in all cases to work out where it should go and how it should fit within Primer's existing API.

Yes, I'd come to the exact same conclusion, with the exception of the `maximum` and `minimum` changes to `Foreword` that you also independently fixed.

primer-miso/primer-miso.cabal
dhess: Yes, I have a fix for this.

primer-miso/Makefile.wasm32
dhess: > Alternatively we could presumably create an actual Nix output for this? Note that in the absence of Wasm support from Haskell.nix, we just get a `nix run .#primer-miso:exe:primer-miso` that doesn't actually do anything. This is because when we compile natively, we currently use a `main` function which only really works in GHCI...

I don't think so, because the Wasm compilers we currently get via `ghc-meta-wasm` aren't hooked into either `haskell.nix` or nixpkgs's `haskellPackages`, so there's currently no way to build the dependencies and store them in the Nix store, as would be required for a proper Nix derivation.

dhess: > We could alter our use of Jsaddle to give us a usable native binary, but honestly I'm not sure it's useful.
> 
> What might be useful is moving the comment from `main` in to an actual make target for ease of development. And potentially substituting GHCId with [ghciwatch](https://github.com/mercurytechnologies/ghciwatch), which is meant to be a modern replacement.

Using `make` for the dev target sounds like a good idea, so I'll add that.

Even though the native executable isn't useful, I still want to build it (via Nix) just to catch any issues in the Nix CI builds.

dhess: > Maybe we should add some instructions to the README for this at least? It wasn't immediately obvious that the command to serve the compiled Wasm is `nix develop .#wasm -c make -f Makefile.wasm32 serve-frontend`.

Yes, I just haven't gotten around to this yet.

primer-miso/Makefile.wasm32
```diff
@@ -0,0 +1,26 @@
+# NOTE:
+#
+# This Makefile assumes you're using the `nix develop .#wasm` shell.
+
+CABAL = wasm32-wasi-cabal
+
+build:	configure
```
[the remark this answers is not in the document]
dhess: Ahh, I was wondering where that came from. That seems to be something new, because the existing Makefiles have been around for over a year and have worked fine until recently.

georgefst:
> Bummer, I have my own fixes for most of this stuff that I worked on during a long flight, but hadn't yet committed to GitHub. Oh well.

Ah, that's a shame to have duplicated work, although I'd be interested to see what you did. My contributions here are mostly pretty trivial apart from my solution to the HLint `maximum` warning, which I'd be surprised if you fixed in exactly the same way.

primer-miso/Makefile.wasm32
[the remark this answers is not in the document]
georgefst: I don't see anything here that should have changed the behaviour. I figured I hadn't personally hit this until now because I'd never even used the Wasm shell.

dhess:
Your changes look nearly identical to mine, except that your new `hlint` warnings about `maximum` and `minimum` are better than mine, as you bothered to find the problematic module prefixes (`Data.Foldable.minimum`).

georgefst:
> Your changes look nearly identical to mine, except that your new `hlint` warnings about `maximum` and `minimum` are better than mine, as you bothered to find the problematic module prefixes (`Data.Foldable.minimum`).

Ah fair enough! That was fairly involved.

primer-miso/build-frontend.sh
georgefst: Oh, weird. Maybe I should double check the behaviour on my end. I wouldn't have expected this to be an OS difference.

primer-miso/Makefile.wasm32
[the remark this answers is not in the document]
dhess: Fixed.

primer-miso/exe/Main.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE CPP #-}
+
+#ifdef wasi_HOST_OS
+
+module MyMain (main) where
+
+import Foreword
+
+--import GHC.Wasm.Prim
```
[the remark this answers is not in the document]
dhess: Done.

primer-miso/Makefile.wasm32
dhess: I've added some instructions to the `wasm.md` documentation on how to serve the new frontend. Once the new Wasm frontend is good enough to replace the TypeScript version, I'll update all the docs (including the top-level README) to reflect the new development workflows.

primer-miso/primer-miso.cabal
dhess: Fixed.

primer-miso/Makefile.wasm32
dhess: Is this breakage a consequence of breaking `primer-miso` out into a separate lib and exe? If so, and you think it's critical to the development workflow, I guess we could undo that. (However, I'd probably want to do that as a separate PR after this one, as I'd like there to be some record of why we did it.)

dhess:
OK, I think modulo a squash and possibly any changes we want to make to accommodate hot reloading (but I prefer to do that in a separate PR, see https://github.com/hackworthltd/primer/pull/1304#discussion_r1870965526), I think this is ready to merge.

dhess:
The build is breaking now because the `pretty-simple` branch disappeared, I assume because of this:

https://github.com/cdepillabout/pretty-simple/pull/107

dhess:
We'll need to wait for haskell.nix to update its snapshot, then bump `index-state`, then rebase, so let's hold off on merging this for a few more days.

primer-miso/Makefile.wasm32
georgefst: > Is this breakage a consequence of splitting primer-miso out into a separate lib and exe?

Yes, but let's keep all this as it is for now, while I continue investigating what can be fixed upstream.

> i.e., using `wasm32-wasi-cabal` rather than native `cabal`

To be honest, I'm surprised it even got close to working with `wasm32-wasi-cabal`. I'm guessing it got further than your native `cabal` attempt because the latter used an older version of Cabal?

> However, I confess I'm not sure what the benefit of this is vs just recompiling, which is pretty fast.

It's the difference between (very roughly) a 10s feedback loop and 0.5s, so it's a gamechanger for experimenting with certain kinds of changes. Admittedly, that 10s is much lower than it would have been with GHCJS, and I don't know whether `jsaddle-warp` would have been developed otherwise. The longer-term plan is some sort of GHCI-in-browser workflow, but I suspect that's still a few releases away.

primer-miso/build-frontend.sh
georgefst: I'm now failing to reproduce this, so let's assume it's not a problem (although I don't currently have access to the same machine on which I originally had the issue).

georgefst (approved):
Looks good.

I'm ambivalent on squashing. The history is now messier than what we usually aim for, but I don't think it's _too_ bad. And, being pedantic, there are changes here that are fairly or even completely orthogonal to `primer-miso` (HLint and Foreword changes, removal of `configure` from Makefiles...).

dhess:
OK. My main concern is broken builds on some of the commits, but those are only due to warnings, so they'll be easy enough to deal with when bisecting.

PR #1305, opened 2024-11-25
Tell Fourmolu about Foreword operator fixities

georgefst:
Basically reverts #1158. I wasn't really actively involved with Fourmolu at the time that Ormolu overhauled its operator handling, or else I would have noticed that the huge diff in that PR was due to an awkward interaction between that new feature and custom preludes (the release notes could have drawn attention to this, and I intend to improve the documentation myself when I get a chance).

### DIFF
--- a/fourmolu.yaml
+++ b/fourmolu.yaml
@@ -8,3 +8,15 @@ haddock-style: single-line
 newlines-between-decls: 1
 single-constraint-parens: auto
 
+reexports:
+  - module Foreword exports Protolude
+fixities:
+  # The `optics` composition operator.
+  # We have no way of telling Fourmolu that Foreword hides the same-named operator from Protolude.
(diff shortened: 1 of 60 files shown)


dhess:
Can you please add `style:` to the head of each of these commit messages? This helps us keep track of which commits are pure style and can be ignored when chasing bugs or looking for places where functionality changed.

dhess (changes requested):
Personally, I like the leading `$` as opposed to the ragged approach, but it's not conventional, so this is fine.

Please change the commit messages to match our conventional commits style, though, and then we're good to merge.

georgefst:
Ah yes, I'd forgotten about Conventional Commits after not working on this project for a while. My contributions to #1304 have the same problem...

dhess (approved):

PR #1308, opened 2024-12-10
chore: bump Nix dependencies and `index-state`

dhess:
Also, bump minimum version of `pretty-simple` to `4.1.3.0` and remove our Wasm workaround repo (which no longer exists).


dhess:
(We had a few known generative test failures in the Wasm merge queue build, but I don't think they're related to the version bumps.)

PR #1312, opened 2025-01-02
chore(primer-selda): build for Wasm

dhess:
Note that the `primer-selda` tests build, but fail, because they try to run `primer-sqitch` in a separate process, but `process` throws a runtime exception under Wasm (on purpose):

https://github.com/haskell/process/blob/166971ffd5a750dd0b101f208f2ddc624d8d6997/System/Process.hs#L904

This is completely orthogonal to whichever Sqlite implementation we choose, of course, and nothing to do with Selda. In fact, in theory, Selda supports its own migration system, so we may be able to replace `primer-sqitch` with Selda's own functionality. (That said, I believe Selda's support for migrations is not particularly robust.)

In any case, if we want to use Sqlite in Wasm, this probably means the end of the road for `sqitch`, and moving the database init/migration into native Haskell code.

### DIFF
--- a/Makefile.wasm32
+++ b/Makefile.wasm32
@@ -15,6 +15,7 @@ test-targets = discover-tests test
 $(test-targets):
 	$(MAKE) -C primer $@
 	$(MAKE) -C primer-api $@
+	$(MAKE) -C primer-selda $@
 
 update:
 	$(CABAL) update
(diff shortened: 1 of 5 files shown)


dhess:
We're probably switching to Beam, so I'm closing this.

PR #1315, opened 2025-01-07
chore: bump dependencies

dhess:


dhess:
This started off as an attempt to build the project with GHC 9.12.1. The Primer stuff builds fine, but the tools in our Nix shell need a lot of work and aren't yet ready for 9.12.1, especially HLS. Therefore, I'll prematurely wrap this PR up with just a Hackage pin bump.

Issue #1307, opened 2024-12-05
Do we need the special `wasm` Nix develop shell?


dhess:
I *think* we can just add the Wasm tools to the default shell and everything will work. I'll test this.

dhess:
Turns out this is probably more trouble than it's worth, mainly due to the fact that `haskell.nix` has some nice built-ins for building dev shells for its projects, and they don't appear to compose nicely with `nixpkgs`'s own tools for creating dev shells.

Issue #1240, opened 2024-04-15
`tasty_two_interp_agree` property test failure


dhess:
```
two interp agree:                               FAIL (4.73s)
--
  | ✗ two interp agree failed at test/Tests/EvalFullInterp.hs:431:60
  | after 212 tests, 24 shrinks and 187 discards.
  | shrink path: 212/187:dCaDeA3fCbA2dGaEaBa2Ba2
  | both terminated         99% ███████████████████▊
  | one failed to terminate  0% ····················
  |  
  | ┏━━ test/Tests/Eval/Utils.hs ━━━
  | 51 ┃ genDirTm :: PropertyT WT (Dir, Expr, Type' () ())
  | 52 ┃ genDirTm = do
  | 53 ┃   dir <- forAllT $ Gen.element @[] [Chk, Syn]
  | ┃   │ Syn
  | 54 ┃   (t', ty) <- case dir of
  | 55 ┃     Chk -> do
  | 56 ┃       ty' <- forAllT $ genWTType $ KType ()
  | 57 ┃       t' <- forAllT $ genChk ty'
  | 58 ┃       pure (t', ty')
  | 59 ┃     Syn -> forAllT genSyn
  | ┃     │ ( App
  | ┃     │     ()
  | ┃     │     (App
  | ┃     │        ()
  | ┃     │        (EmptyHole ())
  | ┃     │        (Case
  | ┃     │           ()
  | ┃     │           (Letrec
  | ┃     │              ()
  | ┃     │              LocalName { unLocalName = "x" }
  | ┃     │              (EmptyHole ())
  | ┃     │              (TEmptyHole ())
  | ┃     │              (Hole
  | ┃     │                 ()
  | ┃     │                 (Case
  | ┃     │                    ()
  | ┃     │                    (Ann
  | ┃     │                       ()
  | ┃     │                       (Letrec
  | ┃     │                          ()
  | ┃     │                          LocalName { unLocalName = "x" }
  | ┃     │                          (EmptyHole ())
  | ┃     │                          (TEmptyHole ())
  | ┃     │                          (Con
  | ┃     │                             ()
  | ┃     │                             GlobalName
  | ┃     │                               { qualifiedModule = ModuleName { unModuleName = "Builtins" :\| [] }
  | ┃     │                               , baseName = "False"
  | ┃     │                               }
  | ┃     │                             []))
  | ┃     │                       (TCon
  | ┃     │                          ()
  | ┃     │                          GlobalName
[...]

PR #1323, opened 2025-02-24
Improve development workflow (`ghcid` with `jsaddle-warp`)

georgefst:
@dhess Can you let me know whether `make develop-frontend` works for you? I think I've solved all the issues from https://github.com/hackworthltd/primer/pull/1304#discussion_r1861223840, but I haven't tested on MacOS.

### DIFF
--- a/primer-miso/Makefile
+++ b/primer-miso/Makefile
@@ -4,5 +4,11 @@
 
 include ../makefiles/common.mk
 
+# TODO The last line here shouldn't be necessary.
+# For some reason, GHCID is failing to discover the right paths to watch.
 develop-frontend:
-	ghcid -c "cabal repl exe:primer-miso" -W -T ':main'
+	ghcid \
+		-c "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" \
+		-W -T Main.main \
+		--reload frontend/style.css \
+		--reload exe --reload src --ignore-loaded \
--- a/primer-miso/Makefile
+++ b/primer-miso/Makefile
@@ -3,3 +3,6 @@
 # This Makefile assumes you're using the `nix develop` shell.
 
 include ../makefiles/common.mk
+
+develop-frontend:
+	ghcid -c "cabal repl exe:primer-miso" -W -T ':main'
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ openapi.json: build
 	cabal run -v0 primer-service:exe:primer-openapi > $@
 	openapi-generator-cli validate --recommend -i $@
 
+develop-frontend:
+	$(MAKE) -C primer-miso develop-frontend
+
 clean:
 	cabal clean
 	rm -f openapi.json
--- a/flake.nix
+++ b/flake.nix
@@ -366,6 +366,7 @@
               ghc982Tools = final.haskell-nix.tools "ghc982" {
                 hlint = "latest";
                 cabal-fmt = "latest";
+                ghcid = "latest";
               };
 
               sqitch = final.callPackage ./nix/pkgs/sqitch {
@@ -487,8 +488,6 @@
                   withHoogle = true;
 
                   tools = {
-                    ghcid = "latest";
-
                     haskell-language-server.src = pkgs.haskell-nix.sources."hls-2.9";
 
                     implicit-hie = "latest";
@@ -517,6 +516,7 @@
 
                     hlint
                     cabal-fmt
+                    ghcid
 
                     # For Language Server support.
                     nodejs-18_x
@@ -671,7 +671,7 @@
               inherit (benchmarks) primer-criterion-results-github-action-benchmark;
               inherit (benchmarks) primer-benchmark-results-github-action-benchmark;
 
-              inherit (ghc982Tools) cabal-fmt hlint;
+              inherit (ghc982Tools) cabal-fmt hlint ghcid;
             }
           );
 
--- a/primer-miso/exe/Main.hs
+++ b/primer-miso/exe/Main.hs
@@ -36,7 +36,7 @@ import Primer.Miso (start)
 main :: IO ()
 main =
   debugOr
-    8000
+    8001
     (insertStylesheet "style.css" >> start)
     (staticApp $ defaultWebAppSettings "frontend")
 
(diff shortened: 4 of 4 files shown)


primer-miso/Makefile
```diff
@@ -4,5 +4,11 @@
 
 include ../makefiles/common.mk
 
+# TODO The last line here shouldn't be necessary.
+# For some reason, GHCID is failing to discover the right paths to watch.
```
georgefst, line 8: I tried to look in to why we need to manually specify the paths. I think it's to do with using multiple components, but I still couldn't reproduce the issue in a simpler project. Anyway, I'd prefer to move to Ghciwatch eventually, but it doesn't yet support multiple components at all: https://github.com/MercuryTechnologies/ghciwatch/issues/316.

georgefst: Also, I considered adding `lib:primer` and `--reload ../primer/src`, in order to also pick up changes to the core library. But this brings the minimum reload time from about 0.5s on my machine to around 3s, even when the only changes are in the frontend components.

Without any of these issues, the command could be as simple as `ghcid -c "cabal repl all" -Wr --reload frontend/style.css`.

dhess:
Is `make develop-frontend` meant to be run in the `nix develop` shell, or the `nix develop .#wasm` shell? I assume the latter, but in either case, I get an error on macOS:

```
$ nix develop
Restored session: Mon Feb 24 16:01:07 WET 2025
bash-5.2$ make develop-frontend
make -C primer-miso develop-frontend
make[1]: Entering directory '/Users/dhess/git/primer/primer-miso'
ghcid \
	-c "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" \
	-W -T Main.main \
	--reload frontend/style.css \
	--reload exe --reload src --ignore-loaded \

Loading cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso ...
Error: [Cabal-5490]
Cannot find the program 'ghc'. User-specified path '/nix/store/z712k8z69c4h2c8303j30f2x6ims36mr-wasm32-wasi-ghc-9.10/bin/wasm32-wasi-ghc' does not refer to an executable and the program is not on the system path.

Command "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" exited unexpectedly with error message: 
make[1]: *** [Makefile:10: develop-frontend] Error 1
make[1]: Leaving directory '/Users/dhess/git/primer/primer-miso'
make: *** [Makefile:28: develop-frontend] Error 2

```

```
$  nix develop .#wasm
evaluation warning: Explicitly set the ABI version of 'webkitgtk'
bash-5.2$ make develop-frontend
make -C primer-miso develop-frontend
make[1]: Entering directory '/Users/dhess/git/primer/primer-miso'
ghcid \
	-c "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" \
	-W -T Main.main \
	--reload frontend/style.css \
	--reload exe --reload src --ignore-loaded \

Loading cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso ...
Error: [Cabal-5490]
Cannot find the program 'ghc'. User-specified path '/nix/store/z712k8z69c4h2c8303j30f2x6ims36mr-wasm32-wasi-ghc-9.10/bin/wasm32-wasi-ghc' does not refer to an executable and the program is not on the system path.

Command "cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso" exited unexpectedly with error message: 
make[1]: *** [Makefile:10: develop-frontend] Error 1
make[1]: Leaving directory '/Users/dhess/git/primer/primer-miso'
make: *** [Makefile:28: develop-frontend] Error 2

```

dhess:
On a related note, I tried to unify the shells so that we only need one, as I understand it's a bit of a pain, but until `haskell.nix` supports Wasm, I don't think it's worth the major refactoring that would be required: see https://github.com/hackworthltd/primer/issues/1307

georgefst:
> Is `make develop-frontend` meant to be run in the `nix develop` shell, or the `nix develop .#wasm` shell? I assume the latter

It's actually the former. This workflow doesn't use Wasm at all.

> , but in either case, I get an error on macOS:

I suspect you've got a `cabal.project.local` file hanging around that you need to delete. The output you pasted shows the non-Wasm shell is still trying to use the Wasm GHC, which is very bad. See https://github.com/hackworthltd/primer/pull/1304#discussion_r1857392029.

georgefst:
> I ran `cabal update`, which isn't something I'd normally expect to need when using a `haskell.nix` shell, but perhaps there's no way around it?

Hmm. I don't know how to explain that. I didn't have to do it, but then I regularly use a non-Nix Haskell toolchain anyway and regularly update my index.

> Anyway, I am able to load the editor in the browser now, which is cool!

That's odd as well, given that it looks like GHCI has crashed. Plus it's not clear why it's crashed. What happens when you just run `cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso`?

> I guess this means we can have native Primer in the browser, as well as Wasm?

Sort of. With `jsaddle-warp` the Haskell code still runs natively outside of the browser, but uses websockets to tell the frontend what JS to run (or something like that - this isn't really documented anywhere AFAICT). Anyway, it's a bit flaky, and annoyingly doesn't work in Firefox, but it's useful for us until the Wasm backend has better GHCI support.

dhess:
> > Anyway, I am able to load the editor in the browser now, which is cool!
> 
> That's odd as well, given that it looks like GHCI has crashed. Plus it's not clear why it's crashed. What happens when you just run `cabal repl --enable-multi-repl exe:primer-miso lib:primer-miso`?

Sorry, I see that my earlier comment was confusing.

The error message I showed is what I got before I ran `cabal update`. After I ran `cabal update`, the `make develop-frontend` target built Primer and successfully launched Primer, with interaction via the browser at `http://localhost:8081`. IOW, it's working fine now.


> > I guess this means we can have native Primer in the browser, as well as Wasm?
> 
> Sort of. With `jsaddle-warp` the Haskell code still runs natively outside of the browser, but uses websockets to tell the frontend what JS to run (or something like that - this isn't really documented anywhere AFAICT). Anyway, it's a bit flaky, and annoyingly doesn't work in Firefox, but it's useful for us until the Wasm backend has better GHCI support.

Yeah, it's never exactly been clear to me how this part of the stack works, either. In any case, what I meant by "native Primer in the browser" is that we can run the native executable (at native speeds), but interact with it in the browser, and we don't need a JS/TS frontend.

dhess (approved):

georgefst:
[Property test failure with interpreter](https://buildkite.com/hackworthltd/primer-wasm-ci/builds/235#01953a1a-7d33-4026-aa80-a9ece090c75a). May be worth recording.

dhess:
Yeah, this is a common failure: https://github.com/hackworthltd/primer/issues/1240

I'll merge manually.

PR #1324, opened 2025-02-24
Use experimental Miso components support

georgefst:
See https://github.com/dmjio/miso/pull/766.

This works, until the last commit, where we try to actually add a component, which causes the app to crash with `TypeError: obj.mount is not a function`. The Miso components branch does not yet support Wasm (or `jsaddle-warp`) due to a need for synchronous callbacks, but this is being worked on at this very moment.

If upstream support does appear in the next week or so, then I propose that we use this as the basis for the next stage of frontend development, i.e. we rebase [my WIP actions and eval work](https://github.com/hackworthltd/primer/compare/main...georgefst/frontend-eval) on top of this. Both the actions and eval panels would benefit from being components since they have a natural notion of internal state. I would otherwise have considered fake components as outlined [here](https://github.com/FPtje/miso-component-example), but I'd say we're better off embracing the future. The success of the frontend rewrite so far has been built on using bleeding edge features and working closely with upstream maintainers. I see no reason to stop now!

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -71,13 +71,12 @@ source-repository-package
   subdir: selda selda-sqlite
   --sha256: 0fw336sb03sc54pzmkz6jn989zvbnwnzypb9n0ackprymnvh8mym
 
--- Until a new Hackage release is made which includes
--- https://github.com/dmjio/miso/pull/752
+-- Until a new Hackage release is made which includes components
 source-repository-package
   type: git
   location: https://github.com/dmjio/miso
-  tag: 2b548d48bffb0e8ae28a6cfb886e4afb0d8be37a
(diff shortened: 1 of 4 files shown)


dhess:
We're in no hurry, so we might as well wait a bit for the components Wasm support.

georgefst:
Well, we didn't have to wait very long! Miso's components now work with `jsaddle-warp` and `jsaddle-wasm`. Both use async callbacks for mounting which is unsafe in theory. In the Wasm case, sync callbacks are [coming soon](https://gitlab.haskell.org/haskell-wasm/ghc-wasm-meta/-/tree/wip/wasm-jsffi-sync-export) but I think there's some work needed in `jsaddle` and `miso` for it to filter through.

dhess:
We've already switched to Miso components.

PR #1331, opened 2025-03-27
chore: un-pin ghc-wasm-meta and fix `primer-miso`

dhess:
Note that this also bumps `primer-miso` to the latest (1.8.7.0). 

We did have to un-pin `experimental` and `jsaddle-wasm` to get this working, however. Hopefully that's temporary until we have proper upstream GHC Wasm releases.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -39,7 +39,7 @@ benchmarks: True
 
 tests: True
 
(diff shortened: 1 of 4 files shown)


dhess:
The merge build failed on a known generative test, so I'm merging this anyway.

PR #1333, opened 2025-03-27
chore: remove primer-service

dhess:
We're moving to a GHC-based web frontend implementation, so we don't need primer-service anymore, and as we were the only user of it, we can do this with impunity.

It would be reasonably easy to resurrect this API in the future, at least the Haskell-friendly Servant bits, if we change our minds. However, it's unlikely that we would bring back the OpenAPI bindings, as those required significant effort to maintain, and due to the impoverished JSON-based serialization API, shed much of the type safety and other benefits of developing applications in Haskell. Having said that, we do still mention OpenAPI here and there, and even keep most of our accommodations for it in `primer-api`, in case we or anyone else want to explore that path again in the future.

Note that while I've tried to clean up the documentation to remove references to the service, the HTTP API, etc., it's likely there are still some references and other non sequiturs.


georgefst:
This actually makes a significant amount of the `primer-api` package redundant, e.g. simplified types like `API.Prog` which were designed for non-Haskell clients. I guess these aren't picked up by Weeder because they're used in tests.

PR #1338, opened 2025-05-06
fix: bump `ghc-wasm` flake to fix `wizer`

dhess:
The version of `wizer` pulled in by this updated `ghc-wasm` flake supports tail calls, which we now need.

(This `ghc-wasm` flake version breaks the Wasm build on macOS, or at least it does on macOS 15.4.1, but we'll just have to deal with the breakage for now, as the need for the updated `wizer` overrides a working macOS build.


dhess:
I've added a workaround for the broken macOS Wasm build: just add `--no-sandbox` to any `nix develop` invocations on that platform.

PR #1341, opened 2025-05-12
Add support for editing programs

georgefst:


primer-miso/src/Primer/Miso/Util.hs
```diff
@@ -250,3 +287,89 @@ nodeSelectionType =
           THole{} -> True
           TEmptyHole{} -> True
           _ -> False
+
+-- TODO keep an eye out for whether these ever return `Nothing` in practice
+-- the current hypothesis is that, because we already typecheck on initialisation and on performing each action,
+-- the app only ever deals with typechecked expressions
+-- and that it is therefore possible that we could refactor the library to assert this in the Haskell types
+-- note that we try to always work with typechecked expressions in the frontend,
+-- as being able to assume that metadata is always available makes the core application code simpler
```
dhess, line 295: ~Can you elaborate a bit on this point in a comment here in the PR thread? This point is probably my only real concern with this PR, and I just want to make sure it's obviously called out in the GitHub UI, so that it's easier to find.~

Never mind, I think it's obvious enough now with my subsequent comments. Let's just not mark this comment as "resolved" so that it remains obvious.

dhess: My initial thought is that probably we should just change `primer-api` so that it always returns type-checked exprs, ASTs, etc. Obviously that's out of scope for this PR and will require quite a bit of refactoring, as you say.

primer-miso/src/Primer/Miso/Util.hs
```diff
@@ -250,3 +286,89 @@ nodeSelectionType =
...
+assumeDefSelectionHasTypeCheckInfo =
+  traverseOf #node $ traverse $ traverseOf #meta $ bitraverse (sequenceOf _type) $ bitraverse (sequenceOf _type) pure
+
+-- see `assumeDefHasTypeCheckInfo`
+-- sometimes we need to discard that extra type-level information,
+-- in order to get an input for various Primer library functions
+astDefTtoAstDef :: ASTDefT -> ASTDef
+astDefTtoAstDef def = ASTDef{astDefExpr = exprTtoExpr def.expr, astDefType = typeTtoType def.sig}
+
+-- this is potentially a better API then the one which `Primer.Action.Available` currently exports
```
dhess, line 311: Note to self: look at `Primer.Action.Available` and see if we should refactor this.

primer-miso/src/Primer/Miso/Util.hs
```diff
@@ -250,3 +286,89 @@ nodeSelectionType =
...
+  Just nodeSel -> case nodeSel.nodeType of
+    BodyNode -> Available.forBody tydefs level editable (astDefExpr def) (getID nodeSel)
+    SigNode -> Available.forSig level editable (astDefType def) (getID nodeSel)
+
+-- this part of the actions API needs a re-think
+-- (it was perhaps too motivated by what was convenient for our old TypeScript frontend):
+-- `context` is a bit weird, and basically just means module (lack of context can mean local name or primitive)
+-- `optToName` is used for presented choices - we know these are variables, not chars or ints
+-- use of `unsafeMk` functions is necessary even though we need never convert from names to text in the first place
+-- `matchesType` is meaningless in `stringToOpt` - we should use different types for presented options and submissions
```
dhess, line 332: Can you file an issue for this? I want to make sure these insights aren't lost in the code base.

primer-miso/src/Primer/Miso/Util.hs
```diff
@@ -250,3 +286,89 @@ nodeSelectionType =
...
+optToName opt =
+  maybe
+    (Left . unsafeMkLocalName)
+    (curry $ Right . unsafeMkGlobalName)
+    opt.context
+    opt.option
+stringToOpt :: Text -> Available.Option
+stringToOpt t = Available.Option t Nothing True
+
+-- should DRY with function of some name in `primer-api`
```
dhess, line 343: I think you mean "same name" here. Anyway, probably worth adding to the TODO list.

dhess (approved):

dhess:
This is a lot less code than I'd thought it would be, nice!

PR #1342, opened 2025-05-12
Allow viewing non-editable definitions, and grey them out in the definitions panel

georgefst:


dhess (approved):

dhess:
The changes here are only to `Primer.Miso`, which has no tests, so there's no point in waiting for a full Wasm CI cycle. Merging.

PR #1347, opened 2025-05-21
Improve Miso string performance

georgefst:
I noticed that rendering performance was surprisingly slow with large trees, and that it scaled linearly with the number of `class_` attributes (around 1ms each).

The first commit is really a leftover from a first attempt at solving it. Unfortunately there was no noticeable impact on performance, even with Miso bumped to a version which uses `JSString` on Wasm instead of `Text`. But there's no harm in keeping it around now that the work has been done. It will be required eventually, and would probably become difficult to merge if left on a separate branch.

The second commit actually solves the problem. I'm still investigating exactly why Miso is so slow without this fix.

### DIFF
--- a/primer-miso/src/Primer/Miso.hs
+++ b/primer-miso/src/Primer/Miso.hs
@@ -586,28 +586,29 @@ viewNodeData position dimensions edges node = case node.opts of
           Clay.width $ Clay.px $ realToClay dimensions.x
           Clay.height $ Clay.px $ realToClay dimensions.y
       ]
-  _ -> div_
-    ( [ class_
-      . mconcat
-      . intersperse " "
-      $ [ "node"
-        , case node.level of
-            Expr -> "expr"
-            Type -> "type"
-            Kind -> "kind"
-        , case node.opts of
-            SyntaxNode{} -> "syntax"
-            _ -> "non-syntax"
-        , case node.opts of
-            SyntaxNode{flavor} -> flavor
-            HoleNode{} -> "hole"
-            PrimNode{} -> "prim"
-            ConNode{} -> "con"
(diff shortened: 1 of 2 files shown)


primer-miso/src/Primer/Miso.hs
```diff
@@ -586,32 +586,36 @@ viewNodeData position dimensions edges node = case node.opts of
           Clay.width $ Clay.px $ realToClay dimensions.x
           Clay.height $ Clay.px $ realToClay dimensions.y
       ]
-  _ ->
-    div_
-      ( [ class_ "node"
-        , class_ case node.level of
+  _ -> div_
```
dhess: Should we add a comment here about why this is necessary?

dhess (commented):
The last commit should be `style: reformat`. Also, per conventional commits we don't capitalize the first word of each commit message, but I'm not gonna be a stickler about it.

primer-miso/src/Primer/Miso.hs
georgefst: Yes, probably. I've added to #1348.

georgefst:
> The last commit should be `style: reformat`.

Ah, that conflicts slightly with the style I use in my personal projects, which is "format" for stuff which wasn't previously formatted (usually, as here, in order to keep a previous diff smaller), and "reformat" for when I've done something like adding or removing line breaks. But I'll try to remember to stick to your preference.

> Also, per conventional commits we don't capitalize the first word of each commit message, but I'm not gonna be a stickler about it.

Oh, damn, I'd forgotten that.

dhess:
It's not a huge deal, just nice to remain consistent.

PR #1349, opened 2025-05-21
Add option to render node IDs in frontend for debugging

georgefst:
Brings back a useful feature we had in the old React frontend.

This is based on #1347 for now, since there'd be conflicts otherwise.


primer-miso/src/Primer/Miso.hs
```diff
@@ -110,6 +110,7 @@ import Primer.Core (
   ),
   GVarName,
   GlobalName (baseName, qualifiedModule),
+  ID (..),
```
dhess, line 113: Nice, this will be useful when I look into whether node IDs are stable (in order to use them for animation purposes).

dhess (approved):

PR #1352, opened 2025-06-02
Add UI for editing type definitions

georgefst:
After #1351, ~~which this is based on top of,~~ this is the final major feature for the new frontend. We can now build any program!

UI is again proof-of-concept, but I don't expect to make major changes there for now. In fact, I think that while it's hardly intuitive in its current form (better labels would help for a start, but this is true for much of the current UI), this is the direction we should be heading in. The choice we've made to only show one definition at a time frees us from worrying about showing the whole type definition as connected, which was always a bit weird in the React frontend. So we now get back some of the clarity of the form-based UI we had in our PureScript frontend, but with the ability to actually create arbitrarily-complex type defs (the fully form-based approach was never going to generalise properly, as I argued a lot back in the day).

~~Code needs some fairly minor tidying up, which will be done once #1351 is merged.~~

Two bugs:
- Selecting a kind node in a constructor field type causes the root node to be selected. This must be an issue with the core library, since it's actually present in the old frontend, though I'd never noticed it before now. I am a bit surprised this isn't caught be some property test along the lines of "perform the action to select a node; that node is now selected", but perhaps this is harder to test for than it sounds. I noticed the issue from absent-mindedly clicking around, though it is arguably pretty obscure in practice, since selecting such a node implies that one probably wishes to edit the kind, and therefore is constructing a type definition which involves quantifying over a higher-kinded type variable. Probably the _simplest_ use case for this is defining monad transformers, which I don't think anyone has attempted in Primer. (Triggering this in the old frontend is even more obscure because selection changes aren't sent to the backend, so whatever buggy Haskell code sets the selection isn't triggered until an actual kind edit is performed.)
- Performing actions which require user input (add param, add constructor, or construct forall) can cause the whole app to crash and need a refresh. The console shows `[WASI stdout] A JavaScript exception was thrown! (may not reach Haskell code)` and `[WASI stdout] TypeError: jsaddle_values.get(...) is undefined`. This happens around half the time, in my experience, and I've detected no clear pattern as to when it does. It's really quite an odd bug. The code handling the input form is very simple, with no obvious explanation for how it can be non-deterministic, or why I've never seen the error with non-type definitions. Anyway, I suspect this doesn't stem from the changes in this PR either, and is a library or compiler issue (I can't really find any relevant issues anywhere, but [this](https://github.com/ghcjs/jsaddle/commit/b8e9793455562e3f30f45f9e311443a76ea1dc4f) is the closest). I'd suggest bumping Miso and/or our Wasm GHC to see whether that makes a difference. I haven't seen anything like this on other projects, where I've mostly been using more recent development versions of Miso.

Some examples:
Screenshot From 2025-06-02 11-24-23
Screenshot From 2025-06-02 11-25-02
image

Note that seeing as we don't actually have buttons for adding definitions yet, I had to add this temporary code in order to get one to play around with:
```diff
diff --git a/primer-miso/src/Primer/Miso.hs b/primer-miso/src/Primer/Miso.hs
index 8ddb47a7..0327f592 100644
--- a/primer-miso/src/Primer/Miso.hs
+++ b/primer-miso/src/Primer/Miso.hs
@@ -230,7 +230,7 @@ start =
             }
       , update = updateModel
       , view = viewModel
-      , subs = []
+      , subs = pure \s -> liftIO $ s $ ApplyProgAction $ AddTypeDef (Primer.qualifyName (Primer.mkSimpleModuleName "Main") "T") $ ASTTypeDef [] [] []
       , events = defaultEvents
       , initialAction = NoOp "start"
       , mountPoint = Nothing
@@ -299,10 +299,12 @@ data Action
   | ToggleFullscreenEval
   | ChooseRedex ID
   | StepBackEval
+  | ApplyProgAction ProgAction

 updateModel :: Action -> Model -> Effect Action Model
 updateModel =
   fromTransition . \case
+    ApplyProgAction action -> runMutation (Edit [action]) >> resetActionPanel
     NoOp _ -> pure ()
     SelectDef d -> do
       #readOnlySelection .= Nothing
```

### DIFF
--- a/primer-miso/src/Primer/Miso/Util.hs
+++ b/primer-miso/src/Primer/Miso/Util.hs
@@ -19,9 +19,14 @@ module Primer.Miso.Util (
   runTC,
   TypeT,
   TermMeta',
+  SelectionT,
+  TypeDefSelectionT,
+  TypeDefNodeSelectionT,
+  DefSelectionT,
   NodeSelectionT,
   ExprMetaT,
   TypeMetaT,
+  ASTTypeDefT,
   KindMetaT,
   ASTDefT (..),
   ModuleT (..),
@@ -30,17 +35,18 @@ module Primer.Miso.Util (
   typeBindingsInExpr,
   bindingsInType,
   nodeSelectionType,
-  DefSelectionT,
   realToClay,
   availableForSelection,
   setSelectionAction,
   astDefTtoAstDef,
+  astTypeDefTtoAstTypeDef,
   optToName,
   stringToOpt,
   assumeDefHasTypeCheckInfo,
-  findASTDef,
+  assumeTypeDefHasTypeCheckInfo,
   assumeDefSelectionHasTypeCheckInfo,
-  selectedDefName,
+  assumeTypeDefSelectionHasTypeCheckInfo,
+  findASTTypeOrTermDef,
   runMutationWithNullDb,
 ) where
 
@@ -76,15 +82,24 @@ import Optics (
   Field2 (_2),
   atraversalVL,
   lensVL,
+  over,
   sequenceOf,
   traverseOf,
+  traversed,
   (%),
   (.~),
   (^.),
+  _Just,
  )
 import Optics.State.Operators ((<<%=))
 import Primer.API (APILog, Env (Env), edit, runPrimerM)
-import Primer.Action (ProgAction (MoveToDef), setCursorBody, setCursorSig)
+import Primer.Action (
+  ProgAction (MoveToDef, MoveToTypeDef, MoveToTypeDefCon, MoveToTypeDefParam),
+  setCursorBody,
+  setCursorSig,
+  setCursorTypeDefConField,
+  setCursorTypeDefParamKind,
+ )
 import Primer.Action.Available (Action)
 import Primer.Action.Available qualified as Available
 import Primer.App (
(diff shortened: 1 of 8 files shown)


dhess:
The UI is obviously extremely basic, but it's probably best to start this way until we figure out what we want.

Do you think that displaying constructors top-down is the right way to go, or was that just the most expeditious implementation?

primer-miso/src/Primer/Miso/Util.hs
```diff
@@ -328,21 +364,41 @@ availableForSelection ::
...
+    Just (TypeDefConsNodeSelection sel') -> case sel'.field of
+      Nothing -> Available.forTypeDefConsNode level editable tydefs defs defName def
+      Just sel'' -> Available.forTypeDefConsFieldNode sel'.con sel''.index (getID sel''.meta) level editable tydefs defs defName def
+    Nothing -> Available.forTypeDef level editable tydefs defs defName def
+  (SelectionDef DefSelection{def = defName, node}, Right def) -> case node of
+    Nothing -> Available.forDef defs level editable defName
+    Just nodeSel -> case nodeSel.nodeType of
+      BodyNode -> Available.forBody tydefs level editable (astDefExpr def) (getID nodeSel)
+      SigNode -> Available.forSig level editable (astDefType def) (getID nodeSel)
+  _ -> [] -- TODO allow raising a warning instead (when def and selection don't match)
```
dhess, line 384: Oof, any thoughts on how to make this block of code less verbose/repetitive?

georgefst: Actually, as alluded to in the comment above this function, this arguably should be the interface for that module, i.e. just take a selection.

PR #1354, opened 2025-06-02
Vendor JS dependencies, for offline support

georgefst:
It's a bit silly that currently the only thing that stops the app from working without an internet connection is downloading a JS shim from some CDN on initial load. I don't even really know what this JS code does: the import line is just copied straight from Tweag's Miso Wasm examples repo.

Vendoring like this isn't really the best option. We should probably download as part of the build process instead. This is really just here as a reminder that we should do _something_ about this, as well as being some code to pull in temporarily at the start of train journeys...


dhess:
Closing, will be fixed in #1429

Issue #1348, opened 2025-05-21
Frontend cleanup


georgefst:
As the new frontend moves from an experiment towards, let's say, alpha quality, we ought to start bringing it up to the standards of the rest of the codebase. This issue serves to collect such concerns in one place. We'll likely solve some of them soon and open separate issues for those which are less trivial.

See https://github.com/hackworthltd/primer/pull/1341#discussion_r2084388749 for background, plus other threads in that PR review as well as https://github.com/hackworthltd/primer/pull/1346.

- [ ] Improve comments.
  - [ ] Add GitHub issues for all `TODO` comments.
  - [ ] Explain Miso `class_` workaround: https://github.com/hackworthltd/primer/pull/1347#discussion_r2100715346.
- [ ] Avoid `error`.
- [ ] Warn on unexpected behaviours.
- [ ] Ideally upstream everything in `Primer.Miso.Util`. At least move the Primer stuff in to `primer` or `primer-api`, and split the rest in to more specific modules or packages.
- [ ] Implement a better form of persistence. Storing the app in local storage with `startComponentWithSavedState` has worked remarkably well, but it was only ever really intended as a bit of a hack to help with a hot-reloading-like workflow.
- [ ] Do some basic profiling and optimisations. I expect there to be a lot of low-hanging fruit here.
  - [ ] For example, currently the app gets pretty sluggish when the number of actions run gets in to the hundreds. This can be mitigated for now by reloading after running `localStorage["miso-app-state"] = JSON.stringify((m => (m.app.currentState.prog.progLog.unlog = [], m.app.initialState = m.app.currentState, m))(JSON.parse(localStorage["miso-app-state"])))` in the browser console, though obviously this breaks the ability to undo older actions.
- [ ] Split some code in to separate modules. `Miso.hs` is now 1000 lines long, making HLS and live reloading noticeably sluggish. The obvious way to split would be by component, in which case we should perhaps wait for the next release of Miso. EDIT: or use HEAD: https://github.com/hackworthltd/primer/pull/1357
- [ ] There are various small missing features, which are obvious from using the app for a few minutes. @dhess has expressed in interest in implementing some of these as easy first contributions to the new frontend app. The list below is not intended to be exhaustive.
  - [ ] Add buttons for adding and removing definitions, including type defs.
  - [ ] Grey out undo and redo buttons when we're at the beginning or end of history respectively.
[...]

PR #1357, opened 2025-06-30
Use Miso HEAD

georgefst:
This has more robust Wasm support, seems to fix the second bug mentioned in https://github.com/hackworthltd/primer/pull/1352, and unblocks component support (mentioned in https://github.com/hackworthltd/primer/issues/1348).


dhess:
I'll fix the formatting checks (they're just Nix formatting failures and won't affect the build) and then merge this manually.

(The macOS test failure is one of the usual suspects and not a new issue.)

PR #1404, opened 2025-12-27
Develop with Wasm GHCI browser mode instead of JSaddle

georgefst:
[See docs](https://downloads.haskell.org/ghc/latest/docs/users_guide/wasm.html#using-ghci-with-the-wasm-backend).

The stuff in the last commit doesn't work yet for a few reasons. GHCIWatch doesn't work with multiple components (https://github.com/MercuryTechnologies/ghciwatch/issues/316), although I'm close to a workaround for this. Note that GHCID doesn't work at all with the Wasm GHCI browser mode, as discovered a few days ago on the Miso Matrix channel. And anyway, Wasm browser mode doesn't work with multiple components in the first place, as I've just discovered and also raised on Matrix.

Since I collapsed the frontend components together in an earlier commit, `make -f Makefile.wasm32 frontend-watch` _does_ work, and allows us to get reload-on-save behaviour for changes in `primer-miso`. However, this is not much use until the browser mode [supports hosting static files](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/15251), because the Primer app is pretty unusable without its CSS. In the meantime, it'd probably be good to split `Miso.hs` in order to speed up reloads.


georgefst:
> So do I understand the intent of this PR correctly in that we would drop native in-browser development altogether? (I am perfectly happy to do this.)

Yes. If it weren't for HLS, we could in theory drop native GHC completely at this point in favour of Wasm.

> Should we go back to compiled-in CSS with Clay/Miso CSS?

Maybe, yeah. In my current non-Primer experiments, this is now the one thing I'm unsatisfied with. Since browser mode also doesn't work with TH, I've ended up doing things like copying a whole CSS file in to a Haskell string...

dhess:
Today I was daydreaming about a Tailwind replacement written as a type-safe (to the extent that CSS semantics are compatible with type safety, anyway) Haskell DSL. Oh well.

dhess:
Closing, see #1423

PR #1423, opened 2026-03-11
feat: refactor Miso frontend

dhess:
We make the following changes:

- Bump Miso to 1.9 release.

- Take advantage of Miso's recent live GHCi reloading support. Note that as a result, we've temporarily dropped our own support for saving Primer sessions across browser reloads until we can refactor it for the new live reloading.

- Drop support for native compilation/debugging/runing the Miso frontend application. The Primer Miso code still builds using GHC native code backends, but as Miso no longer uses JSaddle, it's not particularly practical to run the frontend natively. We only plan to support Primer's UI in browser environments, anyway, so this is no great loss.

Note that the various ways to scaffold live GHCi reloading with Miso in the browser while using the Wasm backend is very much a work in progress, and some functionality may not (always) work as expected until this scaffolding stabilizes.


dhess:
Build failures here are due to the GHC GitLab flaking out. I'm running the Wasm tests locally, and if those pass, I'll merge manually.

In this merge, only `make -f Makefile.wasm32 serve-frontend` will work properly. I'll follow up with another PR where we take some lessons from https://github.com/georgefst/quadris, and perhaps that will fix the remaining debugging modes.

PR #1434, opened 2026-04-27
fix(wasm): drop Brotli compression

### LINKED ISSUE #1433: wasm: drop Brotli compression
We're deploying to Cloudflare now, and our Brotli-compressed `app.wasm` file appears to be causing problems. From the Chrome developer tools console:

```
(index):1 Uncaught CompileError: WebAssembly.instantiateStreaming(): expected magic word 00 61 73 6d, found 9d f0 11 41 @+0
```

Interestingly, the same thing happens with Chrome from a locally-served Brotli-compressed `app.wasm`, whereas it works fine in Safari. (Safari also chokes on the Cloudflare-served version, however.)

Let's just drop Brotli and let our CDN do the compression work.

dhess:
Fixes https://github.com/hackworthltd/primer/issues/1433


dhess:
Something odd is going on with Linux (Kubernetes) CI, where it occasionally rebuilds GHC instead of hitting the cache. This may have something to do with IOG's cache refusing requests from us because we can launch so many builds in parallel now, I don't know, but it is pretty annoying.

I'll merge this manually as there are clearly no code changes here.

